PR #3, opened 2016-03-10
Parse parameter decorators

shuhei:
Because [Method Parameter Decorators](https://goo.gl/8MmCMG) is now [a TC39 stage 0 proposal](https://github.com/tc39/ecma262/blob/master/stage0.md), I'd like to propose babylon to parse parameter decorators.

While the actual transform is still being debated, it would be great if babylon only parses the syntax and allows users to transform it with third-party plugins. One concrete use case is Angular 2's decorators.

https://phabricator.babeljs.io/T1301


test/fixtures/experimental/decorators/parameter-decorator/expected.json
```diff
@@ -0,0 +1,388 @@
...
+                    },
+                    "end": {
+                      "line": 2,
+                      "column": 22
+                    }
+                  },
+                  "name": "x",
+                  "decorators": [
+                    {
+                      "type": "Decorator",
```
thejameskyle: `Decorator` should probably be broken into an alias of many different types.

```ts
interface Decorator <: Node {
  expression: Expression;
}

interface ClassDecorator <: Decorator {
  type: "ClassDecorator"
}

interface ParameterDecorator <: Decorator {
  type: "ParameterDecorator"
}

// etc...
```

loganfsmyth:
This was auto-closed because this repo has been updated to include the Git history from the main Babel repo. It doesn't actually look like Github will let me re-open it? I can't speak to the validity of this PR, but it appears you'll need to rebase and open a new one.

shuhei:
@loganfsmyth Thanks, I'll `cherry-pick` and reopen it!

PR #6, opened 2016-03-12
Fix flow plugin when flow+arrow+spread used together

danez:
Fix for 
```
( ...props: SomeType ) : ?ReturnType => ( 3 );
```

The fix creates a new method on the parser called `parseArrow`.
This new function by default only checks if current position matches an
arrow. If it does returns the `node` otherwise `undefined`.
The flow plugin can then extend this function and correctly parse the typeAnnotation
and add it to the node.

With this change, in the flow plugin there is no need anymore to extend 
`parseParenAndDistinguishExpression` and the arrow handling in `parseParenItem` 
could also be removed, because it is all handled now in `parseArrow`.

Some existing tests were failing, because `extra->parentesized` is now missing, 
but this should be correct as it is now inline with the parsing without flow annotations. No `extra` 
is added for arrow function without type annotations.

Also, in the expression-parser `this.next()` was replaced by a more specific 
`this.expect(tt.parenL)` inside `parseParenAndDistinguishExpression`.

And I reverted the fix from babel/babel#3305 as it is not necessary anymore.

Fixes babel/T2612


loganfsmyth:
This was auto-closed because this repo has been updated to include the Git history from the main Babel repo. I can't speak to the validity of this PR, but it appears you'll need to rebase and open a new one.

PR #12, opened 2016-03-21
Parse parameter decorators

shuhei:
Because [Method Parameter Decorators](https://goo.gl/8MmCMG) is now [a TC39 stage 0 proposal](https://github.com/tc39/ecma262/blob/master/stage0.md), I'd like to propose babylon to parse parameter decorators.

While the actual transform is still being debated, it would be great if babylon only parses the syntax and allows users to transform it with third-party plugins. One concrete use case is Angular 2's decorators.

https://phabricator.babeljs.io/T1301

(Reopening #3 that was automatically closed because of rebased `master`)

### DIFF
--- a/src/parser/lval.js
+++ b/src/parser/lval.js
@@ -162,7 +162,14 @@ pp.parseBindingList = function (close, allowEmpty, allowTrailingComma) {
       this.expect(close);
       break;
     } else {
+      let decorators = [];
+      while (this.match(tt.at)) {
+        decorators.push(this.parseDecorator());
+      }
       let left = this.parseMaybeDefault();
+      if (decorators.length) {
+        left.decorators = decorators;
+      }
       this.parseAssignableListItemTypes(left);
       elts.push(this.parseMaybeDefault(null, null, left));
     }
--- a/test/fixtures/experimental/decorators/class-method-parameter/actual.js
+++ b/test/fixtures/experimental/decorators/class-method-parameter/actual.js
@@ -1,0 +1,3 @@
+class Foo {
+  constructor(@foo() x, @bar({ a: 123 }) @baz() y) {}
+}
--- a/test/fixtures/experimental/decorators/class-method-parameter/expected.json
+++ b/test/fixtures/experimental/decorators/class-method-parameter/expected.json
@@ -1,0 +1,388 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 67,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 3,
+      "column": 1
+    }
+  },
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 67,
+    "loc": {
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
+        "line": 3,
+        "column": 1
+      }
+    },
+    "sourceType": "script",
+    "body": [
+      {
+        "type": "ClassDeclaration",
+        "start": 0,
+        "end": 67,
+        "loc": {
+          "start": {
+            "line": 1,
+            "column": 0
+          },
+          "end": {
+            "line": 3,
+            "column": 1
+          }
+        },
+        "id": {
+          "type": "Identifier",
+          "start": 6,
+          "end": 9,
+          "loc": {
+            "start": {
+              "line": 1,
+              "column": 6
+            },
+            "end": {
+              "line": 1,
+              "column": 9
+            }
+          },
+          "name": "Foo"
+        },
+        "superClass": null,
+        "body": {
+          "type": "ClassBody",
+          "start": 10,
+          "end": 67,
+          "loc": {
+            "start": {
+              "line": 1,
+              "column": 10
+            },
+            "end": {
+              "line": 3,
+              "column": 1
+            }
+          },
+          "body": [
+            {
+              "type": "ClassMethod",
+              "start": 14,
+              "end": 65,
+              "loc": {
+                "start": {
+                  "line": 2,
+                  "column": 2
+                },
+                "end": {
+                  "line": 2,
+                  "column": 53
+                }
+              },
+              "computed": false,
+              "key": {
+                "type": "Identifier",
+                "start": 14,
+                "end": 25,
+                "loc": {
+                  "start": {
+                    "line": 2,
+                    "column": 2
+                  },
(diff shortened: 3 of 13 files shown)


test/fixtures/experimental/decorators/parameter-decorator/expected.json
```diff
@@ -0,0 +1,388 @@
...
+                    },
+                    "end": {
+                      "line": 2,
+                      "column": 22
+                    }
+                  },
+                  "name": "x",
+                  "decorators": [
+                    {
+                      "type": "Decorator",
```
shuhei: @thejameskyle commented here:

https://github.com/babel/babylon/pull/3#discussion_r56099178

shuhei: @thejameskyle Thanks for your comment!

I agree that different types of decorators should be renamed and have `Decorator` as an alias to allow transformers to exactly match desired nodes. However, as far as I looked through [babel-types](https://github.com/babel/babel/tree/master/packages/babel-types), there are already 6 different types of decorators that are all expressed as `Decorator`:

- ObjectMethod
- ObjectProperty
- ClassDeclaration
- ClassExpression
- ClassMethod
- ClassProperty

I think the renaming can be done separately from this PR because it will involve a lot of changes to other packages like babel-types.

thejameskyle: I'm worried how this will impact existing transforms. Adding a `Decorator` type in a new location will cause transforms that are looking for existing types of decorators without validating the rest of the AST will break.

Maybe this one can be the first Decorator node of a new sub-type and we can update the rest later?

shuhei:
@thejameskyle I've updated the code to use `ParameterDecorator` type as you suggested.

Also, I've excluded function parameter decorators because [the proposal](https://docs.google.com/document/d/1Qpkqf_8NzAwfD8LdnqPjXAQ2wwh8BBUGynhn-ZlCWT0/edit#) says:

> ### Function expression and function declaration parameters
>
>This proposal currently targets only method parameters, because function expressions and declarations don't currently have a proposal that would be in Stage 0 or higher.

IgorMinar:
Function expression decorators and function parameter decorators have been
excluded from the proposal due to semantical issues (primarily the temporal
dead zone issues).

It wouldn't hurt to have syntactical support for these two so that we can
experiment with various implementations to get the semantics right.

I do expect that over time, especially if the implementation based on
mirrors is picked as the winner, function expression decorators and
function parameter decorators will be included in the proposals.

shuhei:
@IgorMinar I'd like to parse function parameter decorators too but I'd rather want this PR to be merged sooner.

@thejameskyle What do you think about supporting function parameters decorators at this time?

IgorMinar:
I'm fine with either. If excluding function declaration and function
parameter decorators makes it easier to get this merged then go ahead
without those two. We primarily care about function expression decorators
and method parameter decorators.

thejameskyle:
It's probably fine if they are be behind separate [`this.hasPlugin("...")`](https://github.com/babel/babylon/search?utf8=%E2%9C%93&q=hasPlugin) checks, which the work in this PR should already be behind.

kittens:
Sorry I haven't had a chance to look at this sooner. This looks good besides the fact that `ParameterDecorator` should just be `Decorator`. We already overload this node type for class decorators. We can look into splitting it up in a major but I'd prefer we stick to just `Decorator` for consistency.

shuhei:
@kittens @thejameskyle Updated `ParameterDecorator` to `Decorator` and dropped the distinction between method and function parameters. Please let me know if there's anything to be done before it's merged.

@IgorMinar Just to make it clear, function expression decorators are currently not supported due to [this check](https://github.com/babel/babylon/blob/v6.7.0/src/parser/statement.js#L157-L159).

hzoo:
Don't know if we want a test for params in other node type that are functions: `ArrowFunctionExpression`, `ExportDefaultDeclaration`?

`method-parameter-decorator` is `ClassMethod` in babylon, and there's also `ObjectMethod`.

shuhei:
@hzoo Thanks for your advice. I'll add tests for the other node types.

shuhei:
@hzoo I added tests for other node types except `ArrowFunctionExpression`. It does not work and I'd like to leave it as is.

`ArrowFunctionExpression`'s parameter decorator throws `"Leading decorators must be attached to a class declaration"` error in `parseDecorators` with this PR's implementation. `ArrowFunctionExpression`'s parameter is parsed by `parseParenAndDistinguishExpression` (with parens) or `parseExprAtom` (without parens) without knowing that it is a parameter until `=>` appears. In the parse methods, `parseDecorators` throws the error because the parameter is not a class. The other node types' work because they are parsed by `parseBindingList` that doesn't use `parseDecorators`.

To support `ArrowFunctionExpression`'s parameter decorators, we have to strip the class-only check in `parseDecorators` or parse `ArrowFunctionExpression`'s parameters differently somehow. Both ways might affect the current behavior of Babylon and I don't think they are worth doing while `ArrowFunctionExpression`'s parameter decorator is not actually in the stage 0 proposal yet.

hzoo:
Sounds good, I didn't think arrow functions were in there but I didn't check the proposal or anything :stuck_out_tongue:.

IgorMinar:
@hzoo @thejameskyle is this PR good? Can we get it merged? thanks!

hzoo:
:+1: from me

hzoo:
@kittens 

We may want to figure out the issue with runtime 5/6 before we release I think? And we haven't made a release yet since moving the repo

shuhei:
Thanks everyone :tada:

@hzoo Do we need to update other packages before releasing it? I'm afraid that a new property `decorators` on existing node types may be invalidated by node type/property check.

hzoo:
Yeah we'll need to update babel-types at least https://github.com/babel/babel/tree/master/packages/babel-types/src/definitions, but the checks are restrictive checks so it shouldn't error? (Benefits of keeping it in the monorepo are showing :smile:). 

The decorators transform isn't working atm and there's logan's https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy

shuhei:
@hzoo Thanks! As far as I know, the types that will get `decorators` as a field will be:

- Identifier
- AssignmentPattern
- ArrayPattern
- ObjectPattern
- RestElement

The legacy transformer shouldn't error because it doesn't use `Decorator` visitor.

PR #14, opened 2016-03-24
fixes #13: limit decorator use to CallExpression and IdentifierReference

### LINKED ISSUE #13: Unable to decorate generator functions
With the latest babylon@6.7.0 I am unable to decorate generator functions, i.e.

```
class Foo
{
    @deco
    * gen() {}
}
```

will result in an unexpected error.

Here, it will fail if isGenerator is true.

I am currently working on a patch for this and will report back.

RATIONALE

According to 14.3/14.4 of the ECMAScript specification, a GeneratorMethod (14.4) is a valid MethodDefinition (14.3). And in the [decorator specification](https://github.com/wycats/javascript-decorators/#grammar) also states that it can be used with a MethodDefinition and makes no limitations whatsoever.

silkentrance:
This is an initial stab at #13 that realizes the proposal https://github.com/wycats/javascript-decorators/issues/66

What the patch does is replace the call to parseMaybeAssign() with parseExprSubscripts(), in addition it will also eat the first '@' as parseDecorator() gets called by parseExprOps().

Subsequently, CallExpression and IdentifierReference are realized with the benefit of being able to use subscripts, e.g. `@decorators[Symbol](...)` etc.

Not sure about the template handling also available in parseSubscripts(), though.

My initial integration tests using a real project seem to work fine. Needs additional work on the test cases though, see test/fixtures/experimental/uncategorized/63.

Also, the decorator must be made aware of the fact that it is decorating a generator function a/o an async function. But this is part of the transpiler is it not?

How do I create the expected.json?


hzoo:
If it parses successfully it will generate a expected.json for you when you run the test if it isn't provided

silkentrance:
@hzoo thanks for pointing this out. I will check the fixtures and see whether the file is there.

silkentrance:
@hzoo I didn't realize it but the expected.json was generated and seems to be fine.

silkentrance:
@hzoo I have added additional test cases for testing "fringe" use cases of decorators and also invalid uses such as ``@a?b:c``.

silkentrance:
@hzoo @wycats How does or may ``parseExprSubscripts()`` fit into the current [decorator specificaiton proposal](https://github.com/wycats/javascript-decorators) and how can it fit into the [alternate proposal I made](https://github.com/wycats/javascript-decorators/issues/66)?

kittens:
I'm concerned that this would be a breaking change, we'll need to work out how to coordinate it.

hzoo:
covered by https://github.com/babel/babylon/pull/590

PR #17, opened 2016-04-05
Parse for-await statements when asyncGenerators plugin is active

zenparsing:
Per March 2016 TC39 meeting, [async generators and for-await](https://github.com/tc39/proposal-async-iteration) are now at stage 2.  Babylon already supports parsing of async generator functions.  This commit adds support for for-await statements.

For-await statements look like this:

```js
async function f() {
  for await (let x of y);
}
```

I've chosen to output a new node type: "ForAwaitStatement".  Another approach would be to output a "ForOfStatement" with an "await" property set to `true`.

Thanks!


kittens:
Awesome, thank you!

PR #20, opened 2016-04-11
Add `tokenize` to the public API.

davidaurelio:
`tokenize` returns a function that will return the next token for each call, and `null` after tokenizing the source completely.


kittens:
Discussed this with @davidaurelio back when this PR was created but forgot to comment. Tokenising like this wont be reliable as the tokeniser uses some parser state to disambiguate a lot of things. The existing way to do this would be to do `babylon.parse("code").tokens;`

PR #21, opened 2016-04-17
Test babel with dev babylon

danez:
I created a Makefile that can checkout babel and hook the local version of babylon into babel. Then it is possible to run the babel tests with the current local version of babylon.

just run `make bootstrap-babel` once and then `make test-babel`

This is not necessary the ultimate solution, but at least a start to detect regressions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+build
 lib
 node_modules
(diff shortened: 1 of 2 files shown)


hzoo:
👍  Something like this sounds good

PR #24, opened 2016-04-25
Adjust cloning of nodes in import/export-specifiers and obj-destructuring

danez:
The cloning caused comments that where added to the original node to be persisted into the cloned node. And this later on causes problems in babel-eslint/eslint see babel/babel-eslint#163
Espree/Acorn does not have any cloning, that's why it is working there.

This change adds a parameter to omit comments when cloning, as removing the cloning completely
causes tests in babel to fail.


hzoo:
👍

kittens:
Can we make `__clone` always exclude comment properties?

danez:
Ok I changed it so that `__clone` is now always excluding comments.

PR #25, opened 2016-04-27
Support defaults in Flow's type parameter declarations

gabelevi:
The primary goal of this commit is to add the ability to parse type parameter
declarations with defaults, like `type Foo<T = string> = T`. While I was in the
code, I fixed a few small things, like

* Type parameter declarations need 1 or more type parameters.
* The existential type `*` is not a valid type parameter.
* The existential type `*` is a primary type
* The param list for type parameter declarations now consists of
  `TypeParameter` nodes

### DIFF
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -190,28 +190,46 @@ pp.flowParseTypeAlias = function (node) {
 
 // Type annotations
 
+pp.flowParseTypeParameter = function () {
+  let node = this.startNode();
+
+  let variance;
+  if (this.match(tt.plusMin)) {
+    if (this.state.value === "+") {
+      variance = "plus";
+    } else if (this.state.value === "-") {
+      variance = "minus";
+    }
+    this.eat(tt.plusMin);
+  }
+
+  let ident = this.flowParseTypeAnnotatableIdentifier(false, false);
+  node.name = ident.name;
+  node.variance = variance;
+  node.bound = ident.typeAnnotation;
+
+  if (this.match(tt.eq)) {
+    this.eat(tt.eq);
+    node.default = this.flowParseType ();
+  }
+
+  return this.finishNode(node, "TypeParameter");
+};
+
 pp.flowParseTypeParameterDeclaration = function () {
   let node = this.startNode();
   node.params = [];
 
   this.expectRelational("<");
-  while (!this.isRelational(">")) {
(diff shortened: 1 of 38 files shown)


gabelevi:
@kittens or @hzoo - Is there anything I can do to help get this PR merged?

kittens:
@gabelevi Sorry for the delay. Were there any new node types/properties introduced that we necessitate changes to the code generator and traversal definitions in Babel?

gabelevi:
@kittens - sorry, I'm not quite understanding what you're asking. Are you asking about the downstream effects of this PR and what will have to change? I don't think I know enough about the code generator and traversal definitions to comment. Naively, a transform to strip away flow types shouldn't be affected, since this is a change within the type space. However a general visitor might notice that the list of type params is now an array of `TypeParameter` rather than an array of `Identifier`

glenjamin:
This PR appears to fix https://phabricator.babeljs.io/T7365 as well. I had started to have a go at this but only got as far as expanding a test case to cover more existential scenarios before discovering it already worked on `master`.

It might be worth including these examples to cover the newly expanding existential support? (I added the flow header so that I could run flow on the testcase to check it agreed with this parser)

```patch
diff --git a/test/fixtures/flow/type-annotations/existential-type-param/actual.js b/test/fixtures/flow/type-annotations/existential-type-param/actual.js
index 83f6d6f..a10100a 100644
--- a/test/fixtures/flow/type-annotations/existential-type-param/actual.js
+++ b/test/fixtures/flow/type-annotations/existential-type-param/actual.js
@@ -1 +1,8 @@
-type Maybe<T> = _Maybe<T, *>;
+/* @flow */
+type Maybe<T> = Map<T, *>;
+class ABC extends Array<*> {}
+function abc(a: *): number {
+  return 1;
+}
+var obj: {a: *, b: *} = {a: 1, b: "a"};
+var x: * = 17;
```

hzoo:
This is causing an issue in some of the babel tests https://github.com/babel/babel/pull/3519 (a readme change). I guess one of the benefits with keeping it in the same repo was knowing it a change in babylon affected babel without having to guess? We should probably do something like https://github.com/babel/babylon/pull/21/files or automatically run a test with new babylon changes?

hzoo:
WIP PR https://github.com/babel/babel/pull/3520 - and yeah actually these are technically a breaking change (at this specifically the one for `type params is now an array of TypeParameter rather than an array of Identifier` @gabelevi

PR #29, opened 2016-05-07
Add conditional assignment plugin

doug-wade:
Adds a conditional assignment operator

```
let foo;
foo ||= 'foo';
foo === 'foo'; // true

let foo = 'foo';
foo ||= 'bar';
foo === 'foo'; // true
```

I'm not really sure how close this PR is to being ready to pull? I probably need to update the documentation. There's an [associated babel pr](https://github.com/babel/babel/pull/3499).

### DIFF
--- a/src/tokenizer/index.js
+++ b/src/tokenizer/index.js
@@ -319,11 +319,22 @@ export default class Tokenizer {
     return this.finishOp(type, width);
   }
 
-  readToken_pipe_amp(code) { // '|&'
+  readToken_pipe() { // '|'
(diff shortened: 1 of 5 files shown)


jmm:
Hi @doug-wade. Thanks for your interest in contributing to Babylon. As noted in your other PR, generally only standards track stuff is going into the core Babel libraries, with some legacy / pseudo-standard exceptions like JSX. For that reason I'm closing this PR. There's a notion of making Babylon extendable in a way that would allow creation of plugins for handling new syntax like this. You can track status of that at https://phabricator.babeljs.io/T1351. Thanks!

Issue #40, opened 2016-06-12
Document deviations from ESTree spec


nene:
I tried using Babylon parser as a replacement for Esprima, discovering that the AST it produces doesn't quite correspond to the [ESTree spec](https://github.com/estree/estree). Which is even more surprising because the README states that Babylon is based on acorn, which does conform to ESTree spec.

I found no documentation about this. So far I've discovered the following differences:

- [Literal][] token is replaced with [StringLiteral][], [NumericLiteral][], [BooleanLiteral][], [NullLiteral][], [RegExpLiteral][]
- [Property][] token is replaced with [ObjectProperty][] and [ObjectMethod][]
- [MethodDefinition][] is replaced with [ClassMethod][]
- what else?

I'd like to evaluate whether it's feasible for me to switch the parser to Babylon... but I'm not really sure how big the AST differences turn out to be.

[Literal]: https://github.com/estree/estree/blob/master/spec.md#literal
[Property]: https://github.com/estree/estree/blob/master/spec.md#property
[MethodDefinition]: https://github.com/estree/estree/blob/master/es6.md#methoddefinition

[StringLiteral]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#stringliteral
[NumericLiteral]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#numericliteral
[BooleanLiteral]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#booleanliteral
[NullLiteral]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#nullliteral
[RegExpLiteral]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#regexpliteral
[ObjectProperty]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#objectproperty
[ObjectMethod]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#objectmethod
[ClassMethod]: https://github.com/babel/babel/blob/master/doc/ast/spec.md#classmethod

loganfsmyth:
Not really a direct answer, but FYI Babel's AST format is documented here: https://github.com/babel/babel/blob/master/doc/ast/spec.md. We should probably move that to this repo though.

Babel's format is based on Acorn, but does not attempt to correspond directly with ESTree.

nene:
Thanks for the reference.

[...]

PR #41, opened 2016-06-18
Document AST differences from ESTree

nene:
Based on comments in #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,43 @@ Significant diversions are expected to occur in the future such as streaming, EB
 
 - **plugins**: Array containing the plugins that you want to enable.
 
+### Output
+
+Babylon generates AST according to [Babel AST format][].
+It is based on [ESTree spec][] with the following deviations:
+
+- [Literal][] token is replaced with [StringLiteral][], [NumericLiteral][], [BooleanLiteral][], [NullLiteral][], [RegExpLiteral][]
+- [Property][] token is replaced with [ObjectProperty][] and [ObjectMethod][]
+- [MethodDefinition][] is replaced with [ClassMethod][]
+- [Program][] and [BlockStatement][] contain additional `directives` field with [Directive][] and [DirectiveLiteral][]
+
+AST for JSX code is based on [Facebook JSX AST][] with the addition of one node type:
+
+- `JSXText`
+
+[Babel AST format]: https://github.com/babel/babel/blob/master/doc/ast/spec.md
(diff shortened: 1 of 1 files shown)


jmm:
@nene Thanks for the effort on this. I'm definitely +1 on noting the AST output format and linking to the spec. I don't think the README is the ideal place to detail differences from ESTree (or whatever AST spec) though. I think that should be in a separate document, probably not linked from the README -- maybe from the Babel AST spec. That document should maybe be a wiki, since I don't know that anyone maintaining Babel / Babylon is going to maintain that information.

If there is going to be a *differences from ESTree* document (especially if it's committed somewhere) I think we should say something like *differences include* so that we're not setting the expectation that it's exhaustive or up to date. Even ESTree isn't static -- they make additive changes.

hzoo:
The differences were also in the Babel 6 changelog (the AST only changed in 6.0)

nene:
Currently the README tells me that Babylon is based on Acorn (which does follow ESTree), so I would assume Babylon also follows ESTree spec unless told otherwise.

Given that README describes input parameters, I'd expect that it also describes the output.

jmm:
Babylon is indeed based on Acorn, but it's a fork and one of the reasons for forking was to be able to diverge from ESTree. I totally agree that we should say what the output format is and link to the AST spec, I just don't think we should inline too much information in the README, like a breakdown of the differences from ESTree. If there's going to be a breakdown we can link to it from the AST spec or somewhere fitting. I'm going to propose the following:

1. We decide in which repo the AST spec should live. In #40 @loganfsmyth mentioned the possibility of moving it from the `babel` repo to here. I've thought about that before but been on the fence about it. Right now I lean toward keeping it where it is since multiple core packages under the Babel umbrella and in the `babel` repo utilize that AST format.

1. Whichever repo that is, we consider creating a wiki there to break down the differences from ESTree. Alternatively perhaps someone who wants to maintain it creates a repo / gist / whatever for it in userland and we link to it from the AST spec or wherever. I'm not real enthusiastic about the idea of committing it to one of the Babel repos because I'm doubtful that anyone will want to or actually maintain it or that there's even going to be a lot of users interested in it.

1. We update this README to say that the output is *Babel AST* format and link to the spec.

kittens:
We should move the AST docs from the `babel/babel` repo into here. This section in the README is really good though, thank you @nene! Super appreciated

hzoo:
Made https://github.com/babel/babylon/pull/46

PR #42, opened 2016-06-19
Add JSX spread children

calebmer:
This PR comes from a conversation in https://github.com/facebook/jsx/issues/57 where I recommended adding the ability to spread children. For the full rationale see that issue, here’s an abridged version:

* * *

JSX with React allows us to do this:

```jsx
const TodoList = ({ todos }) => (
  <div>
    {todos.map(todo => <Todo key={todo.id} todo={todo}/>)}
  </div>
)
```

Which translates to this:

```js
const TodoList = ({ todos }) => ({
  type: 'div',
  children: [
    todos.map(todo => …)
  ],
})
```

That doesn’t make sense and it has disadvantages for simpler JSX consumers as well as type systems. Instead JSX should allow us to do this:

```jsx
const TodoList = ({ todos }) => (
  <div>
    {...todos.map(todo => <Todo key={todo.id} todo={todo}/>)}
  </div>
)
```

Which would translate to this:

```js
const TodoList = ({ todos }) => ({
  type: 'div',
  children: [
    ...todos.map(todo => …)
  ],
})
```

This PR aims to extend JSX by allowing the spreading of children like so:

```jsx
<div>{...children}</div>
```

* * *

As another note, it seems like React itself will consider this an anti-pattern. However as JSX is an open standard used by more than just React we should consider the entire ecosystem when considering the merits of this extension. Let’s keep discussion of whether or not the extension should be added to JSX in https://github.com/facebook/jsx/issues/57 and let’s talk about technical implementations here.

In https://github.com/facebook/jsx/issues/57 the language extension got a +1 from @syranide (https://github.com/facebook/jsx/issues/57#issuecomment-226773478) and @sebmarkbage (https://github.com/facebook/jsx/issues/57#issuecomment-226843120).

### DIFF
--- a/src/plugins/jsx/index.js
+++ b/src/plugins/jsx/index.js
@@ -345,6 +345,15 @@ pp.jsxParseElementAt = function(startPos, startLoc) {
           break;
 
         case tt.braceL:
+          if (this.lookahead().type === tt.ellipsis) {
+            let node = this.startNode();
+            this.next();
+            this.next();
+            node.expression = this.parseExpression();
+            this.expect(tt.braceR);
+            children.push(this.finishNode(node, "JSXSpreadChild"));
+            break;
+          }
           children.push(this.jsxParseExpressionContainer());
           break;
 
--- a/test/fixtures/jsx/basic/21/actual.js
+++ b/test/fixtures/jsx/basic/21/actual.js
@@ -1,0 +1,1 @@
+<div {...c}> {...children}{a}{...b}</div>
--- a/test/fixtures/jsx/basic/21/expected.json
+++ b/test/fixtures/jsx/basic/21/expected.json
@@ -1,0 +1,273 @@
+{
+  "type": "File",
+  "start": 0,
+  "end": 41,
+  "loc": {
+    "start": {
+      "line": 1,
+      "column": 0
+    },
+    "end": {
+      "line": 1,
+      "column": 41
+    }
+  },
+  "program": {
+    "type": "Program",
+    "start": 0,
+    "end": 41,
+    "loc": {
+      "start": {
+        "line": 1,
+        "column": 0
+      },
+      "end": {
+        "line": 1,
+        "column": 41
+      }
+    },
+    "sourceType": "script",
+    "body": [
+      {
+        "type": "ExpressionStatement",
+        "start": 0,
+        "end": 41,
+        "loc": {
+          "start": {
+            "line": 1,
+            "column": 0
+          },
+          "end": {
+            "line": 1,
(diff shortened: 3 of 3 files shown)


calebmer:
I’m not sure why `jsx/basic/10` is failing. It looks like the comment blocks array at `leadingComments/closingElement/expression/0/body/program` has two identical copies the following comment block instead of the one tested for:

```json
{
  "type": "CommentBlock",
  "value": " this is a comment ",
  "start": 4,
  "end": 27,
  "loc": {
    "start": {
      "line": 1,
      "column": 4
    },
    "end": {
      "line": 1,
      "column": 27
    }
  }
}
```

I’m pretty sure it has to do with the [`lookahead`](https://github.com/babel/babylon/pull/42/files#diff-c77b5c915712a9e45567a2f1e8b863b1R348), but I couldn’t find any documentation on the matter.

kittens:
We'll have to hold off on this until it's been specced and added to the facebook/jsx#57 repo. TypeScript and Flow also support JSX so there's value in having this specced there so they can come into compliance as we add support.

sebmarkbage:
I don't think either TypeScript or Flow support non-React variants of JSX atm so it probably doesn't matter but we can add it. We also don't support namespaces which is speced.

calebmer:
@kittens: The language extension this PR implements was just added to the JSX specification (https://github.com/facebook/jsx/pull/59). What’s remaining to get this PR merged, and can you comment on the failing test?

calebmer:
Just curious @kittens, what’s the status on this?

src/plugins/jsx/index.js
```diff
@@ -345,6 +345,15 @@ pp.jsxParseElementAt = function(startPos, startLoc) {
           break;
 
         case tt.braceL:
+          if (this.lookahead().type === tt.ellipsis) {
+            let node = this.startNode();
+            this.next();
+            this.next();
+            node.expression = this.parseExpression();
+            this.expect(tt.braceR);
+            children.push(this.finishNode(node, "JSXSpreadChild"));
```
danez: `JSXSpreadChild` would need to be added to [babel-types](https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/jsx.js) as well and also to [babel-generator](https://github.com/babel/babel/blob/master/packages/babel-generator/src/generators/jsx.js)

calebmer: :+1: https://github.com/babel/babel/pull/3575

kittens:
Thanks! We can follow up in the Babel PR.

Issue #7, opened 2016-03-14
Add missing `tools/generate-identifier-regex.js`


mathiasbynens:
https://github.com/babel/babylon/blob/a1f42e7ac4239204a35f85d5f9555358ca1c7789/src/util/identifier.js#L37 mentions `tools/generate-identifier-regex.js` but no such script exists in the repo.

It appears this was copied from https://github.com/ternjs/acorn/blob/3b4ab11ead6eb5fc4f9d2a8ef535d3b2ce699af1/src/identifier.js#L35.

Either way, please include the script.

forivall:
I also have a slightly modified version at https://github.com/forivall/tacoscript/blob/master/packages/horchata/bin/generate-identifier-regex.js

PR #45, opened 2016-06-22
Update to Unicode v9

mathiasbynens:
Ref. #7.


kittens:
Thank you @mathiasbynens! I wish I understood more about text encoding.

PR #52, opened 2016-06-22
add-module-exports for backwards compat on package with babel 5

hzoo:


kittens:
Are you sure this is it? We don't export a default from this package

hzoo:
Hmm.. maybe not then. Weird since there shouldn't of been any major changes. We changed from babel 5 to babel 6 in babylon first though (as well as not using loose mode)

PR #54, opened 2016-06-23
[Flow] Arrow function type parameter declarations

gabelevi:
In Flow, you can declare type parameters for arrow functions. The syntax looks like this:

```JavaScript
  const arrowFunctionWithoutTypeParams = (x) => x;
  const arrowFunctionWithTypeParams = <T>(x: T): T => x;
```

This is a little tricky to parse, especially with JSX. This PR attempts to solve the problem as follows

1. Try to parse the expression as if it were JSX. JSX is probably more common and this will lead to less backtracking. If this fails, save the error. If the JSX plugin isn't in use, then skip this step.
2. If the next token is `<`, then parse a type parameter declaration. Then parse an assignment expression. If that expression turns out to an ArrowFunctionExpression, then attach those type params to that expression. Otherwise, fail. On failure throw the JSX error, if it exists, or a new error.
3. If this is not JSX And doesn't start with `<` then just parse an assignment expression as usual.

This was further complicated by the need to reinterpret tokens and mess with context. But at least it seems to work :)


gabelevi:
This fixes https://phabricator.babeljs.io/T7330

src/plugins/flow.js
```diff
@@ -1053,6 +1062,68 @@ export default function (instance) {
...
+  //    fails, we'll try an arrow function. If that fails, we'll throw the JSX
+  //    error.
+  // 2. This is an arrow function. We'll parse the type parameter declaration,
+  //    parse the rest, make sure the rest is an arrow function, and go from
+  //    there
+  // 3. This is neither. Just call the inner function
+  instance.extend("parseMaybeAssign", function (inner) {
+    return function (...args) {
+      let jsxError = null;
+      let state = this.state.clone();
```
kittens: This isn't super obvious but cloning state is actually super expensive, for example there's a `tokens` array on there that gets cloned and this can sometimes number in the tens of thousands for really large files.

kittens: Oh great, just noticed that `state` is only used within the JSX `if`, you can just move it into there. It should resolve the more pathological cases but the perf issue will still be there. We should really use babylon to use immutable data structures so this state cloning is really cheap.

gabelevi: Sure, I'll move it in there. It will still be hit for every JSX expression, I'm afraid, but that's better than every expression :)

kittens:
Thanks @gabelevi! I'll publish a new release now.

hzoo:
We aren't running the changes on babel in travis yet right?

kittens:
@hzoo Nope.

kittens:
I pushed a new release just now, I'll run `make test-babel` to ensure that I haven't broken anything...

hzoo:
Ok cool - it does suck that it would make the tests run a lot longer but worth doing.

PR #55, opened 2016-06-24
flow: allow generic method with name get or set

vkurchatkin:


src/parser/statement.js
```diff
@@ -692,9 +693,13 @@ pp.parseClassBody = function (node) {
     if (!method.computed) {
       let { key } = method;
 
+      if (this.hasPlugin("flow") && this.isRelational("<")) {
+        isGeneric = true;
```
kittens: This logic should be done inside of `plugins/flow.js` via method overloading.

kittens: ie. you could do something like:

```javascript
pp.isClassMutatorStarter = function () {
  return doSomethingHere;
};
```

and use `this.isClassMutatorStarter()` in the `if` check and then overload `isClassMutatorStarter` to do your checking.

vkurchatkin: Ok! I thought about that, but saw a bunch of `hasPlugin` just above

vkurchatkin: Finding a good name is tricky though, so I'll just use `isClassMutatorStarter` for now

kittens:
Awesome, thank you so much @vkurchatkin!

PR #59, opened 2016-06-29
Fix regression with conditional, arrow function and flow types

### LINKED ISSUE #58: Significant but hard-to-find bug in parsing arrow functions
As of a recent release (Don't know when, exactly, but this issue broke code that was working fine around June 4), babylon's parsing of arrow functions has broken in a very peculiar way. When babylon encounters the following pattern:

```
<some expression> :
    <a parenthesized arrow expression> ?
    <an unparenthesized arrow expression>
```

It reports an error parsing the parenthesized arrow expression:

```
Module build failed: SyntaxError: <path to file>: Assigning to rvalue (23:3)
```

For example:

```
let x = foo() ?
    (value => true) :
    value => false
```

The contents of the condition don't seem to matter, nor does parenthesizing the entire conditional expression. However, if the parenthesis around the first arrow function are removed, or parenthesis are added to the second arrow function, or either arrow function is something other than an arrow function, it is parsed fine. This specific pattern is the only one that triggers the error.

For what it's worth, I'm running babylon via babel via webpack, but the stack trace I get is entirely from babylon.

This seems implausible, I know, so I've created a git repo which reproduces the error, available here:

https://github.com/Lucretiel/babylon-error

The specific file I'm trying to pack is [index.js](https://github.com/Lucretiel/babylon-error/blob/master/index.jsx) with no dependencies or requires. The raw output from `webpack` is in [build_result](https://github.com/Lucretiel/babylon-error/blob/master/build_result).

danez:
With 6.8.2 we introduced a regression when arrow functions with a return type are used together with a conditional expression.

The fix is to first try to parse the `consequent` value without evaluating the first arrow function return type we encounter and then checking if the ConditionalExpression is followed by a colon.
If it is (which wouldn't be valid syntax) we know that we interpreted stuff in a wrong way, and try parsing again without the restriction on the arrow function return type.
If the first parse isn't followed by a colon, we can be sure we parsed things correctly.

This logic is always only added to the most outer conditional expression.

Fixes #58


danez:
This approach does not work anymore, need to find another solution

PR #61, opened 2016-07-02
adds indented mode to Jsx via parser option

neurall:
This PR comes from a conversation in facebook/jsx#60 where I proposed adding the ability to use Jade like compact indented JSX syntax via option.

Reasoning.
A lot of Jsx lines are useless end tags. But screen and bandwidth space is precious. And as projects/sources grow ability to focus and spot bugs is often lost in noise especially after inline styles kick in. 

Source changes:
No control flow / state changes. 
all code is enclosed and activated only when indented:true,locations:true passed in parser ptions

Because Jsx parser is simple all that was needed was adding children based on indentation(for now it uses line and column no from (locations:true) in jsxParseElementAt and disable complaining about missing end tags in jsxReadToken and 

All tests pass with and without option. 
I also implemented first indent test in Experimental/Jsx/indented dir along with sample
~~~~~~html
var App  =  <div>
                <drawer> 
                    <menu>
                        <item>
                <area>
                    <toolbar>
                    <content>
~~~~~~
This will probably later result to faster simpler parsing due to half of tags  and smaller ast trees in web
assembly later but most importantly Jsx UI structure will now be much more easy to understand and bugs easier to spot especially in complex projects.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -8,7 +8,8 @@ export const defaultOptions: {
   allowImportExportEverywhere: boolean,
   allowSuperOutsideMethod: boolean,
   plugins: Array<string>,
-  strictMode: any
+  strictMode: any,
+  indented: boolean
 } = {
   // Source type ("script" or "module") for different semantics
   sourceType: "script",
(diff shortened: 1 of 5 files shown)


loganfsmyth:
This would definitely need to be part of the official spec before we're likely to accept it.

neurall:
I definitely understand. I reimplemented it as JadeX plugin as proposed by facebook guys. Thx

PR #63, opened 2016-07-02
Fix performance regression introduced in 6.8.2

### LINKED ISSUE #62: Very slow parsing of large JSON file with flow plugin
With the flow plugin, https://github.com/babel/babylon/pull/19 seems to have regressed parsing performance on really giant files.

Before #19 (e6c11a06), this finished in about 0.5 seconds. After, it takes hours:

```
curl -O https://raw.githubusercontent.com/mbostock/topojson/master/examples/world-50m.json
node -e 'require("/path/to/babylon").parse("(" + require("fs").readFileSync("world-50m.js", "utf-8") + ")", { plugins: [ "flow" ] })
```

There is no problem if the flow plugin is disabled. It looks like the `this.state.clone()` calls get slower and slower as it parses more of the file. Maybe the total runtime is `O(n^3)` in the size of the file? That would put us at about 470 hours to parse the file based on a bit of logging I did.

This is blocking me from updating Facebook to use a newer Babylon. (Actually, the reason I want to upgrade is to get the fix from #19…)

danez:
This commit e6c11a0 (#19) made a big performance regression.
The reason was that parseConditional was always cloning the current state even if no questionmark (potential conditional or flow-optional
token) was at the current position.
Simply checking if questionmark matches the current token solves the problem.

Fixes #62 

Testcase:

Node: v7.0.0-nightly20160621ecc48a154d
Testfile:
https://gist.githubusercontent.com/milafrerichs/69035da4707ea51886eb/raw/4cb1783c2904f52cbb8a258ee96031f9054d155b/eu.topojson

Before:
bildschirmfoto 2016-07-02 um 17 09 47

After:
bildschirmfoto 2016-07-02 um 17 08 18

### DIFF
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -732,18 +732,24 @@ export default function (instance) {
 
   instance.extend("parseConditional", function (inner) {
     return function (expr, noIn, startPos, startLoc, refNeedsArrowPos) {
-      const state = this.state.clone();
-      try {
-        return inner.call(this, expr, noIn, startPos, startLoc);
-      } catch (err) {
-        if (refNeedsArrowPos && err instanceof SyntaxError) {
-          this.state = state;
-          refNeedsArrowPos.start = this.state.start;
-          return expr;
-        } else {
-          throw err;
+      // only do the expensive clone if there is a question mark
+      // and if we come from inside parens
+      if (refNeedsArrowPos && this.match(tt.question)) {
+        const state = this.state.clone();
+        try {
+          return inner.call(this, expr, noIn, startPos, startLoc);
+        } catch (err) {
+          if (err instanceof SyntaxError) {
+            this.state = state;
+            refNeedsArrowPos.start = err.pos || this.state.start;
+            return expr;
+          } else {
+            throw err;
+          }
         }
       }
+
+      return inner.call(this, expr, noIn, startPos, startLoc);
     };
   });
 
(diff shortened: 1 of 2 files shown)


spicyj:
Cool, glad there's a simple fix. I suppose this still could be a problem on files with thousands of question marks, but maybe that's not worth worrying about. Should be fine on JSON-like things which are the most likely contents of a large JS file.

danez:
Yes, I want to run more profiling on babylon, I saw some possible candidates, but I need to check them. Gladly it is really nice, easy and fun to do profiling with chrome dev tools and node 7.

spicyj:
Another thought I just had: if I try to parse `a ? b : c ? d : e ? f : g ? h : i ? j : k ? l : m ? n : o ? p : syntax error`, will it try exponentially many paths when backtracking?

danez:
That might be related to #59. I can check tomorrow.

danez:
Oh sorry I was reading it on the phone yesterday and thought you said that you get an syntax error when parsing nested ternaries. :)

In the given example there would be no backtracking involved now with this PR (i added one more check), as the flow optional operator can only be inside parentheses `(a?)`.
If your example would have parens around every ternary then it would do `n` backtracks where `n = number of nested ternaries` if I'm not mistaken, which should be linear.
Basically the flow plugin is only wrapping the parsing of conditionals into a try-catch-clause and doing a backtrack when a SyntaxError happens, so that other parts of the flow plugin can validate it as valid optional param.

kittens:
Thanks! Published as 6.8.3

spicyj:
Thank you!

PR #64, opened 2016-07-02
Feat/add jadex plugin

neurall:
This PR comes from a conversation in facebook/jsx#60 where I proposed adding the ability to use Jade like compact indented JSX syntax via option. This was proposed to be implemented as separate plugin.
Since Jsx without end tags resembles Jade. I named this plugin JadeX. 

Reasoning.
A lot of Jsx lines are useless end tags. But screen and bandwidth space is precious. And as projects/sources grow ability to focus and spot bugs is often lost in noise especially after inline styles kick in. 

How it works.
All the parsing is still done by Jsx plugin as usual.
What we do is we load jsx before jadex and enable locations collecting. The latter is needed for indentations but will be optimized later with locations only for tagstarts. 
After Jsx extends acorn with his jsx methods.
We extend acorn with our method calling jsxParseIndentedElementAt instead of jsxParseElementAt.
jsxParseIndentedElementAt adds children based on indentation(for now it uses line and column no from (locations:true) 
Finally we disable complaining about missing end tags in jsxReadToken 

All tests pass. 
I also implemented first indent basic test in test/fixtures/Jadex/indented. More will follow.

~~~~~~html
var App  =  <div>
                <drawer> 
                    <menu>
                        <item>
                <area>
                    <toolbar>
                    <content>
~~~~~~
This will also result to twice as fast parsing due to half of tags  and smaller ast trees in web
assembly but most importantly Jsx UI structure will now be much more easy to understand and bugs easier to spot especially in complex projects.


neurall:
Ahh. I work on case insensitive mac so build was passing before PR. 

But as travis spotted
Offender was uppercase plugin dir name 

JadeX instead of jadex.
Sorry for that.

neurall:
i just published in npm babel-plugin-syntax-jadex so babel is now aware of this syntax

var babel  = require("babel-core");
var output = babel.transform("<div>", { presets: ['react'],plugins: ['syntax-jadex'] }).code;

> "React.createElement("div", null);"

loganfsmyth:
Sorry, I guess I could have made this clearer in https://github.com/babel/babylon/pull/61. We're not really looking to add one-off syntax extensions to Babylon. At this point we support JSX and Flow as kind of grandfathered-in extra syntax, but otherwise Babylon is only really aimed at supporting officially proposed features for the ECMAScript specification.

hzoo:
Ok going to close

PR #65, opened 2016-07-03
Fixes SyntaxError position with flow optional type

danez:
After #19 was merged the column that babylon reports for SyntaxErrors involving a question mark was slightly off, because the initial SyntaxError was swallowed, a backtrack performed and then a new SyntaxError generated. For example:

```
var a = (o ? p : =)
```
babylon without flow reports the `=` as unexpected, whereas with flow plugin it reports `p` as unexpected token.

This PR fixes that by forwarding the position.


src/parser/expression.js
```diff
@@ -551,7 +551,7 @@ pp.parseParenAndDistinguishExpression = function (startPos, startLoc, canBeArrow
     if (first) {
       first = false;
     } else {
-      this.expect(tt.comma);
+      this.expect(tt.comma, refNeedsArrowPos.start || null);
```
hzoo: is this just `0` b/c there's ` let refNeedsArrowPos = { start: 0 };` at 550?

danez: basically this is doing: `refNeedsArrowPos.start > 0 ? refNeedsArrowPos.start : null`
This line is first triggered in the second iteration `if (first) ...` and `refNeedsArrowPos` is supplied to `parseMaybeAssign` and might be changed there.

PR #68, opened 2016-07-06
Fix the location of params, when flow and default value used

### LINKED ISSUE #67: Location info for AssignPattern seems to be off when using a type and default parameter
`function g(a:number=1, e=1:number) {}`

for `a:number=1` the `AssignmentPattern` loc start and end is 19-21 [see here](http://astexplorer.net/#/nVQ5xiMwub) when it should be 11-21?

danez:
Fixes #67 and facebook/jscodeshift#127


hzoo:
👍

Issue #71, opened 2016-07-09
Flow declarations don't check for declare name


danez:
Currently declares inside `declare module` don't need to start with a `declare` name. For example this is valid code:
```
declare module A {
  placeanythinghere var a:number
}
```

Flow itself also parses it. I know flow tries not to fail in most situations, but do we also want to do that in babylon?
It could be easily changed and there is already a [todo](https://github.com/babel/babylon/blob/master/src/plugins/flow.js#L101), but I'm not sure what the wanted way currently is.

@jeffmo, @kittens : What are your thoughts on this?

jeffmo:
This is a parse error in Flow -- should probably be in Bablyon, too.

https://astexplorer.net/#/pNxlnaH77f

danez:
Ah, I never checked the errors array. Stupid me.

jeffmo:
Well, I think the other half of this issue is still relevant: Babylon should probably error

danez:
Yes I opened #73

PR #73, opened 2016-07-10
Only allow declares inside declare module

danez:
This is the followup of #71 

This now checks for the term `declare` being used inside `declare module`. Previously everything was allowed instead of declare. https://astexplorer.net/#/Y6hVjJkQ2m


test/fixtures/flow/declare-module/8/options.json
```diff
@@ -0,0 +1,3 @@
+{
+  "throws": "Unexpected token (2:2)"
```
hzoo: Is it possible to have a better error message here?

danez: Maybe something like: "Only declare statements are allowed inside declare module {}"

hzoo: Yeah something like that. Could even give an example (we haven't really done that)

I've been liking elm's errors/messages http://elm-lang.org/blog/compilers-as-assistants

danez:
I changed the error message to be a little bit nicer.

src/parser/util.js
```diff
@@ -79,6 +79,6 @@ pp.expect = function (type) {
 
 // Raise an unexpected token error.
 
-pp.unexpected = function (pos) {
-  this.raise(pos != null ? pos : this.state.start, "Unexpected token");
+pp.unexpected = function (pos, message = "Unexpected token") {
+  this.raise(pos != null ? pos : this.state.start, message);
```
hzoo: I like being able to change the error message. Unexpected token is not fun to see (my example is `function a() { await b; }`) without async in front of function

hzoo:
k 👍 , @jeffmo

gabelevi:
👍 this looks correct to me. Everything inside of a `declare module` declaration does start with `declare`.

PR #75, opened 2016-07-11
Fix incomplete support for exponential operator

danez:
While looking through the es2016+ table [here](http://kangax.github.io/compat-table/es2016plus/) I noticed that babel is not marked green for `exponentiation (**) operator - basic support`. This has to do with the parenthesis that are mandatory for prefixed literals.

So ultimately it should be like this:
```js
5 ** 6; // success
(-5) ** 6; // success
-(5 ** 6); // success
-5 ** 6; // fail
-(5) ** 6; // fail
(-5 ** 6); // fail
```
But currently it is like this:
```js
5 ** 6; // success
(-5) ** 6; // fail
-(5 ** 6); // success
-5 ** 6; // fail
-(5) ** 6; // success
(-5 ** 6); // fail
```

~~I added some currently failng tests.~~


danez:
I found a solution to fix it. Luckily we already have a lot of information about parenthesis in extra.

hzoo:
👍

PR #82, opened 2016-07-26
Fix accidental fall-through in Flow type parsing.

xiemaisi:
When parsing a primary type, `>` would erroneously be treated like `(`. For example, `>x:int) => string` was parsed the same as `(x:int) => string`.

### DIFF
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -506,6 +506,7 @@ pp.flowParsePrimaryType = function () {
 
         return this.finishNode(node, "FunctionTypeAnnotation");
       }
+      break;
 
     case tt.parenL:
       this.next();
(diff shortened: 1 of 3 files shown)


src/plugins/flow.js
```diff
@@ -505,6 +505,8 @@ pp.flowParsePrimaryType = function () {
         node.returnType = this.flowParseType();
 
         return this.finishNode(node, "FunctionTypeAnnotation");
+      } else {
+        this.unexpected();
```
danez: This is just minor and makes not a big difference, but maybe adding here just a `break` and then it would call `this.unexpected()` on line 395

danez:
Good catch. I wonder if we have the same problem with  `tt.plusMin`

xiemaisi:
That certainly also looks suspicious, but I wasn't immediately able to construct an example (I'm new to Babylon).

src/plugins/flow.js
xiemaisi: Sure, that works, too. I've amended the commit.

danez:
Yes I also couldn't figure out an example.
Thanks

PR #94, opened 2016-08-18
Fix arrow functions with destructuring, types & default value

danharper:
e.g. `var x = ({ a } : any = 'foo') => {}`

Will currently throw: "Binding rvalue"

```
repl: Binding rvalue (1:9)
> 1 | var x = ({ a } : any = 'foo') => {}
    |          ^
```

Flow's `toAssignable` override wasn't calling the inner function after setting the type annotation, resulting in the node not having the type clarified from [`ObjectExpression` to `ObjectPattern`](https://github.com/babel/babylon/blob/b6496718a7434a31592d1fce4986b57955bfb6bd/src/parser/lval.js#L21) (which would later cause it to throw as `ObjectExpression` isn't a valid lval).


danez:
Good catch. Thanks for your contribution.

src/plugins/flow.js
```diff
@@ -878,11 +878,11 @@ export default function (instance) {
   });
 
   instance.extend("toAssignable", function (inner) {
-    return function (node) {
+    return function (node, isBinding) {
```
hzoo: We can probably use ...args here and pass that down right? (although I see there's only 2 params)

danharper: That's what I originally had tbh, then noticed `toAssignableList` was being explicit with the args.

Shall I change it?

hzoo: Nah all good 😄

PR #98, opened 2016-08-19
Remove exponentiationOperator, asyncFunctions, trailingFunctionCommas

danez:
This enables them by default now.

Also restructure the test directories, with es2015, es2016, es2017 names ....

I checked that the number of tests stayed the same.


hzoo:
can do the same for `trailingFunctionCommas`, dono if you want that in a separate one

hzoo:
we should deprecate classConstructorCall too

danez:
I also started restructuring and renaming the test dirs
```
-- es2015
-- es2016
```

at first i wanted to do something like

```
--es
  -- 2015
  -- 2016
```

but babel-helper-fixtures is limited to how deep one can nest dirs, so I did the first solution.

Gonna do all in this PR.

danez:
I'm not sure how to proceed with `classConstructorCall` probably we can just remove it with the next major. Should we log a warning if it is used already?

hzoo:
Yeah we would remove it in the next major - just not sure given we would need to update babel too.

PR #101, opened 2016-08-29
Remove deprecated plugin from README.md

danez:
As we probably remove it anyway in the next major we can already at least not expose it in the readme.


hzoo:
It's on the docs site still though (with deprecated)

PR #103, opened 2016-08-31
[Flow] Function predicate declaration

panagosg7:
The accepted syntax for function declarations is extended to allow
the specification of a function predicate (inferred or explicit). This
should be placed right after the function's return type (if existing,
otherwise the colon that would designate the start of the return 
type). So the return type syntax becomes:

```
FunctionReturnType :=
  Type
  Predicate
  Type Predicate

Predicate :=
  %checks
  %checks ( ConditionalExpression )
```

These changes are with regards to [this](https://github.com/facebook/flow/commit/6cbb327b7678a9c7050e7cacaa1ca2cf751c08ae) commit in the Flow repo.

### DIFF
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -375,7 +375,7 @@ pp.flowParseObjectType = function (allowStatic, allowExact) {
       if (variance) {
         this.unexpected(variancePos);
       }
-      nodeStart.callProperties.push(this.flowParseObjectTypeCallProperty(node, allowStatic));
+      nodeStart.callProperties.push(this.flowParseObjectTypeCallProperty(node, isStatic));
     } else {
       propertyKey = this.flowParseObjectPropertyKey();
       if (this.isRelational("<") || this.match(tt.parenL)) {
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -1051,8 +1051,8 @@ export default function (instance) {
   });
 
   // parse type parameters for class methods
-  instance.extend("parseClassMethod", function () {
-    return function (classBody, method, isGenerator, isAsync) {
+  instance.extend("parseClassMethod", function (inner) {
+    return function (classBody, method, ...args) {
       if (method.variance) {
         this.unexpected(method.variancePos);
       }
@@ -1061,8 +1061,8 @@ export default function (instance) {
       if (this.isRelational("<")) {
         method.typeParameters = this.flowParseTypeParameterDeclaration();
       }
-      this.parseMethod(method, isGenerator, isAsync);
-      classBody.body.push(this.finishNode(method, "ClassMethod"));
+
+      inner.call(this, classBody, method, ...args);
     };
   });
 
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -25,10 +25,9 @@ pp.flowParsePredicate = function() {
   this.expectContextual("checks");
   // Force '%' and 'checks' to be adjacent
   if (moduloLoc.line !== checksLoc.line || moduloLoc.column !== checksLoc.column - 1) {
-    this.unexpected(moduloPos);
+    this.raise(moduloPos, "Spaces between ´%´ and ´checks´ are not allowed here.");
   }
-  if (this.match(tt.parenL)) {
-    this.next();
+  if (this.eat(tt.parenL)) {
     node.expression = this.parseExpression();
     this.expect(tt.parenR);
     return this.finishNode(node, "DeclaredPredicate");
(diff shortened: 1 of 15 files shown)


src/plugins/flow.js
```diff
@@ -21,6 +21,45 @@ pp.flowParseTypeInitialiser = function (tok, allowLeadingPipeOrAnd) {
...
+pp.flowParsePredicate = function() {
+  let node = this.startNode();
+  let moduloLoc = this.state.startLoc;
+  let moduloPos = this.state.start;
+  this.expect(tt.modulo);
+  let checksLoc = this.state.startLoc;
+  this.expectContextual("checks");
+  // Force '%' and 'checks' to be adjacent
+  if (moduloLoc.line !== checksLoc.line || moduloLoc.column !== checksLoc.column - 1) {
+    this.raise(moduloPos, "Unexpected token");
```
danez: This line is not covered. Do you think we can have two tests that check if there are parse errors in these cases?

danez:
Thank you for your first PR to babylon. 

In order this works throughout the complete chain of babel, all the new node types (I think it is only DeclaredPredicate  and InferredPredicate) need to be added to [babel-types](https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/flow.js) and [babel-generator](https://github.com/babel/babel/blob/master/packages/babel-generator/src/generators/flow.js)
Could you do that and create a PR for babel?

I would wait with the merge till we have a PR in babel and the new flow version is released.

Great work!

src/plugins/flow.js
danez: You could also use `this.unexpected(moduloPos);` here. It's a little bit shorter but otherwise isn't really important if not.

panagosg7:
Thanks @danez! I'll address your comments

panagosg7:
@danez: This can still wait until Flow's upcoming release.

src/plugins/flow.js
```diff
@@ -21,6 +21,46 @@ pp.flowParseTypeInitialiser = function (tok, allowLeadingPipeOrAnd) {
...
+pp.flowParsePredicate = function() {
+  let node = this.startNode();
+  let moduloLoc = this.state.startLoc;
+  let moduloPos = this.state.start;
+  this.expect(tt.modulo);
+  let checksLoc = this.state.startLoc;
+  this.expectContextual("checks");
+  // Force '%' and 'checks' to be adjacent
+  if (moduloLoc.line !== checksLoc.line || moduloLoc.column !== checksLoc.column - 1) {
+    this.unexpected(moduloPos);
```
danez: Can you add a nice error message here? Maybe something like `"Spaces between ´%´ and ´checks´ are not allowed here."`

src/plugins/flow.js
```diff
@@ -21,6 +21,46 @@ pp.flowParseTypeInitialiser = function (tok, allowLeadingPipeOrAnd) {
...
+  }
+  if (this.match(tt.parenL)) {
+    this.next();
```
danez: Instead of 
```js
if (this.match(tt.parenL)) {
    this.next();
```
you can do 
```js
if (this.eat(tt.parenL)) {
```

Which is doing the exact same thing.

panagosg7:
Thanks for the pointers, @danez!

I will get to these, but I'm still waiting for support for function predicates to be fully merged in Flow master.

loganfsmyth:
I know flow supports this now, is this waiting on more changes, or has it just been forgotten?

src/plugins/flow.js
```diff
@@ -21,6 +21,46 @@ pp.flowParseTypeInitialiser = function (tok, allowLeadingPipeOrAnd) {
...
+    this.expect(tt.parenR);
+    return this.finishNode(node, "DeclaredPredicate");
+  } else {
+    return this.finishNode(node, "InferredPredicate");
+  }
+};
+
+pp.flowParseTypeAndPredicateInitialiser = function () {
+  let oldInType = this.state.inType;
+  this.state.inType = true;
```
panagosg7: It looks like `this.state.inType` needs to be `true` for `this.expect(tt.colon)` to succeed (I tried to move it further down, but other tests failed). So, `oldInType` needs to be recorded in the beginning in order to be restored before `this.flowParsePredicate()` is called (note that `flowParsePredicate` possibly parses expressions).

Bottom line is I wasn't able to perform the optimization you suggested.

panagosg7:
Sorry for taking so long to get back to this.

AFAIK, there is still some function predicate code that hasn't been merged to Flow's master. Most of these changed involve the type checker, but there is also some parser work under way. For example I believe the notation of [predicate refinement](https://github.com/facebook/flow/blob/master/tests/predicates-abstract/filter.js#L5) will be updated to something more elegant. Also, there has not been an official announcement yet.

However, you are right that some support for function predicates (including `%checks`) is already in mainline and recognized by the native Flow parser, so I guess there should be support here as well.

I recently rebased my original PR, addressed @danez 's comments (including the ones in the [respective babel PR](https://github.com/babel/babel/pull/3700)) and added one more example.

leebyron:
Any progress on this? I'd love to see this land to unblock early use

PR #104, opened 2016-09-01
exact object type annotations for Flow plugin

bhosmer:
Adds support for exact object type annotations in the Flow plugin. Exact object types are delimited by `{|` and `|}`, but are otherwise identical in syntax to ordinary object types.

Replicates the functionality added to Flow [here](https://github.com/facebook/flow/commit/c710c40aa2a115435098d6c0dfeaadb023cd39b8).


src/tokenizer/index.js
```diff
@@ -404,8 +405,17 @@ export default class Tokenizer {
...
+      case 123:
+        if (this.hasPlugin("flow") && this.input.charCodeAt(this.state.pos + 1) == 124) {
+          return this.finishOp(tt.braceBarL, 2);
+        } else {
+          ++this.state.pos;
+          return this.finishToken(tt.braceL);
+        }
+
+      case 125:
+      ++this.state.pos; return this.finishToken(tt.braceR);
```
danez: Needs 2 space indentation.

src/plugins/flow.js
```diff
@@ -510,7 +523,8 @@ pp.flowParsePrimaryType = function () {
       return this.flowIdentToTypeAnnotation(startPos, startLoc, node, this.parseIdentifier());
 
     case tt.braceL:
-      return this.flowParseObjectType();
+    case tt.braceBarL:
+      return this.flowParseObjectType(false, true);
```
danez: If the token === tt.braceL shouldn't the call be `this.flowParseObjectType(false, false);`. It probably doesn't really change anything but it would be more clear maybe (and saves us one comparison in the `flowParseObjectType`, although it probably doesn't have an impact.)

danez:
Thank you for bringing the new feature immediately to babylon. We recently are a little bit behind the flow specs with other features, and I'm really happy to see more help in bringing the new features to babylon.

I have two small nitpicks but otherwise looks good. To make the complete chain of babel work with this change there also needs to be a PR to babel that makes [babel-generator](https://github.com/babel/babel/blob/master/packages/babel-generator/src/generators/flow.js#L253) aware of the `exact` field so the syntax can be printed again. Could you make these changes there?

bhosmer:
@danez Sure, will make the requested changes here and in `babel-generator`. Thanks for the quick response Daniel!

bhosmer:
@danez Hi Daniel - incorporated your suggestions. Not sure what to do about the coverage numbers, if anything - I'd have thought the new tests would cover the new code paths, but LMK if there's something else to be added.

Re `babel-generator`, I've got the new code in [my fork](https://github.com/bhosmer/babel/commit/bcce4c05bf9250e3f87ade11bc622deadd8dc40c), but since the new tests choke the old parser, it seems like I should hold off on a PR until this change here makes its way over there, right?

bhosmer:
@danez PR to `babel-generator` is [here](https://github.com/babel/babel/pull/3702).

But I think we may need to bump Babylon's version to reflect the new support, and then specify the new version in `babel-generator`'s dependencies, otherwise the `babel-generator` tests I added which use `{| |}` will fail. Not sure how big a version bump is; alternatively, I could just remove the new tests from the PR. LMK which you'd prefer!

danez:
Coverage looks good and you can create the PR already for babel. The tests will fail there of course but that is okay for features that span over both repos.

danez:
Ah you were faster with your reply ;) yes this is now perfectly fine. We are going to merge the change in babel after babylon gets released.
Thank you

bhosmer:
@danez ah, ok cool - thank you Daniel!

src/tokenizer/index.js
```diff
@@ -404,8 +405,17 @@ export default class Tokenizer {
       case 44: ++this.state.pos; return this.finishToken(tt.comma);
       case 91: ++this.state.pos; return this.finishToken(tt.bracketL);
       case 93: ++this.state.pos; return this.finishToken(tt.bracketR);
-      case 123: ++this.state.pos; return this.finishToken(tt.braceL);
-      case 125: ++this.state.pos; return this.finishToken(tt.braceR);
+
+      case 123:
+        if (this.hasPlugin("flow") && this.input.charCodeAt(this.state.pos + 1) == 124) {
```
thejameskyle: `=== 124`

bhosmer: (y)

thejameskyle:
This is looking good. @hzoo @danez could you push this through?

hzoo:
Looks like the coverage would be for a failing test on `unexpected token`? 

https://github.com/babel/babylon/pull/104/files#diff-8ce9de97e7fb2bd7f30e9eee69714584R406

bhosmer:
@hzoo added test and coverage looks taken care of, I think. Not sure why the build failure though, or how to kick.

hzoo:
I restarted got the same failure - should be ok

bhosmer:
@hzoo cool, thanks!

gabelevi:
@hzoo - thanks for merging! Would you mind publishing this to npm? Thanks!

hzoo:
Sorry been busy not doing open source 😄, can do it real quick

PR #107, opened 2016-09-05
Disallow duplicate named exports

kaicataldo:
I think I got all the cases for named exports - let me know if I missed one! Also wasn't sure about the error message wording, but was hoping that by including the name of the offending duplicate export that debugging could be made easier for the end user. Finally, wasn't sure where to put the tests - happy to move them!


test/fixtures/flow/type-exports/interface/actual.js
```diff
@@ -1,2 +1,2 @@
 export interface foo { p: number };
-export interface foo<T> { p: T };
+export interface bar<T> { p: T };
```
kaicataldo: I'm pretty new to Flow - let me know if this isn't the correct fix for this test

kaicataldo:
Looks like it's failing because of test files in Babel plugins that contain multiple exports in the same file. How is that usually handled? I'm happy to make the necessary changes in the Babel repo too.

danez:
Good job. That looks really good. The flow test looks good. 👍 

The tests of babel now fail, but this is okay for now, but we would need to correct them at some point, before we release this.

kaicataldo:
Cool! Should I go ahead and make a PR to the main babel repo?

danez:
Yes that would be nice. Thanks.

Oh for some reason I did not see the other comment of you. So there is no real workflow yet on how to do the cross-repo stuff. I described in this PR how to link babylon in babel. I should really but that in the Contributing file.

https://github.com/babel/babel/pull/3700#issuecomment-244622489

kaicataldo:
Thanks for all the info on this! One more question - do you think I should split the changes into batches? Splitting each one of those files into multiple files is going to be a pretty massive change.

danez:
Ah sorry, for not answering. As long as the changes belong together I think they should go into the same PR imho. And I think this would account for this case.

kaicataldo:
Finally started in on this, though might take me a few days. Doing it in chunks because there are a lot of changes.

kaicataldo:
Finally got around to fixing those tests! PR open here: https://github.com/babel/babel/pull/4538

I mention it in the issue as well, but will also make a 2nd PR that removes the [initial band-aid fix](https://github.com/babel/babel/pull/3518) that was merged into Babel itself and to rename the test files so they follow the convention used in the currently open PR. I plan on getting to this tomorrow!

src/parser/statement.js
```diff
@@ -844,7 +844,7 @@ pp.parseExport = function (node) {
     }
     node.declaration = expr;
     if (needsSemi) this.semicolon();
-    this.checkExport(node);
+    this.checkExport(node, true, true);
```
hzoo: just a thought: might be nice if this method wasn't taking in true, true (an object, named, etc) but I know we do that everywhere else ^ `this.parseFunction(expr, true, false, false, true);` 😄

kaicataldo: Yeah, was trying to follow the style I saw elsewhere, but agree. Would you like me to change that to an object?

test/fixtures/flow/type-exports/interface/actual.js
danez: This is correct

src/parser/statement.js
danez: @hzoo: Do you mean `checkExport(node: Node, options: { named: boolean, default: boolean })` or what do you mean with object? Maybe also  `checkExport(node: Node, type: "named" | "default" | null)`?

hzoo: yeah to make it more readable - dono if it was done before for perf?

kaicataldo: I assumed it was done this way for performance reasons

kaicataldo:
Opened one more PR that reverts the commit I made that throws a warning in babel itself since we have moved that into babylon. I also updated the test names so that they match the convention I used in my more recent PR that was merged today. Let me know if you need anything else!

dlmr:
Seems like this change causes Babylon to throw for one of my files that worked fine before. 

The file in questions looks like this:
```javascript
export toString from './toString';
```

```bash
`toString` has already been exported. Exported identifiers must be unique. (1:7)
> 1 | export toString from './toString';
    |        ^
  2 |
```
Is this an intended change, is not `toString` allowed any more?

Seems like all keys on `Object.prototype` causes this error.

src/parser/statement.js
```diff
@@ -913,6 +937,20 @@ pp.checkExport = function (node) {
   }
 };
 
+pp.checkDuplicateExports = function(node, name, isDefault) {
+  if (this.state.exportedIdentifiers[name]) {
```
dlmr: I think we need to manage `Object.prototype` here.

Something along the lines of this maybe?

```js
if (Object.getOwnPropertyNames(this.state.exportedIdentifiers).indexOf(name) !== -1) {
```

danez: You are right, I wrap up a fix

hzoo: Ah interesting

cc @kaicataldo

kaicataldo: Oh wow, that makes sense. @danez Let me know if you need me work on a fix - happy to do it.

hzoo: Same issue in https://github.com/jscs-dev/node-jscs/pull/1204 😄

PR #110, opened 2016-09-07
update packages, use es2015 loose mode, remove babel-runtime transform

hzoo:
- update deps
- use flow strip types, remove react preset dev
- remove babel-runtime dep/transform-runtime since we are using loose mode

Should pass on node 0.10


package.json
```diff
@@ -40,8 +38,9 @@
     "lint": "kcheck",
     "prepublish": "cross-env BABEL_ENV=production npm run build",
     "preversion": "npm run test",
-    "test": "npm run build && npm run lint && ava test/",
-    "test_cov": "cross-env BABEL_ENV=test npm run build && npm run lint && nyc ava test/",
+    "test": "npm run build && npm run lint && ava test",
```
leebyron: `&& npm run testonly`?

leebyron:
Can the `babel-runtime` dependency also be removed from package.json?

package.json
hzoo: Oh yeah lol forgot I committed that

hzoo:
@leebyron oh right, that was the whole point 😄

leebyron:
So good!

package.json
```diff
@@ -7,28 +7,24 @@
   "license": "MIT",
   "repository": "https://github.com/babel/babylon",
   "main": "lib/index.js",
-  "dependencies": {
-    "babel-runtime": "^6.0.0"
-  },
+  "dependencies": {},
```
leebyron: you can probably just delete this line

leebyron:
❤️

package.json
```diff
@@ -7,28 +7,24 @@
...
+    "babel-plugin-transform-flow-strip-types": "^6.14.0",
+    "babel-preset-es2015": "^6.14.0",
     "babel-preset-stage-0": "^6.5.0",
     "codecov": "^1.0.1",
-    "cross-env": "^2.0.0",
-    "kcheck": "^2.0.1",
-    "lodash": "^4.6.1",
-    "nyc": "^8.0.0",
+    "cross-env": "^2.0.1",
+    "kcheck": "^2.0.3",
```
danez: should we also use eslint-config-babel instead of kcheck maybe?

hzoo: We can do that in another PR? Can ref https://github.com/babel/babel/pull/3693

hzoo:
And what about running rollup since we don't have deps?

Issue #120, opened 2016-09-11
Parsing support for computed class property names


motiz88:
See Babel issue: babel/babel#4499

hzoo:
https://github.com/babel/babylon/issues/121

PR #121, opened 2016-09-11
Add support for computed class property names (#120)

motiz88:
First time contributor here - someone please check this for sanity :smile:

I _think_ this implements #120 quite completely. I'm working on a PR for the corresponding changes required in Babel, tracked by babel/babel#4499.

### DIFF
--- a/ast/spec.md
+++ b/ast/spec.md
@@ -1021,6 +1021,7 @@ interface ClassProperty <: Node {
   type: "ClassProperty";
   key: Identifier;
   value: Expression;
+  computed: boolean;
 }
 ```
 
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -672,13 +672,13 @@ pp.parseClassBody = function (node) {
       this.parsePropertyName(method);
     }
 
-    if (!isGenerator && method.key.type === "Identifier" && !method.computed) {
+    if (!isGenerator) {
       if (this.isClassProperty()) {
         classBody.body.push(this.parseClassProperty(method));
         continue;
       }
 
-      if (this.hasPlugin("classConstructorCall") && method.key.name === "call" && this.match(tt.name) && this.state.value === "constructor") {
+      if (method.key.type === "Identifier" && !method.computed && this.hasPlugin("classConstructorCall") && method.key.name === "call" && this.match(tt.name) && this.state.value === "constructor") {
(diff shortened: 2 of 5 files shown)


motiz88:
The CI failures (on Node 5 only, out of the whole build matrix) [appear to be unrelated](https://travis-ci.org/babel/babylon/jobs/159109742#L4348) to the PR's content. (Right?)

danez:
Thanks for contributing.

This looks good, the only thing that I'm worried about this is how the ambiguous cases are handled. For example this class:
```
class Bar {
   [x] = blub
   ['y'];
}
```
Does it contain two or one properties? I would assume one with a value of `blub['y']`. But I'm not sure what the spec says about this if it does.
What would your changes produce for this example?

motiz88:
My immediate answer is that ASI applies like everywhere else (and there are existing tests for literal-named properties reflecting this). Doing a little bit of research and will follow up soon.

danez:
I found this
https://github.com/tc39/proposal-class-public-fields/issues/25

Not sure what the current status is, but semicolons might be required in the case i posted.

danharper:
There's some [existing tests](https://github.com/babel/babylon/tree/master/test/fixtures/experimental/class-properties), which cover the ASI rules in this case for computed methods.

danez:
Good catch.

motiz88:
There's also tc39/proposal-class-public-fields/pull/45 (an open PR at the moment) which suggests the semicolon is not in the spec, but IMHO it's safe to assume it will be (given past discussions and the nontrivial ambiguity otherwise).

motiz88:
@danez your `class Bar` parses as having one property.

hzoo:
Yeah ref https://github.com/babel/babel/pull/3332

hzoo:
Looks fine to me? We made semicolons necessary before but it was reverted

jeffmo:
The tests like right to me

danez:
@jeffmo what will the spec say about this case? Or is it not yet decided?

```
class Bar {
   [x] = foo
   ['y'];
}
```

jeffmo:
@danez: Grammatically, that would be the same as:

```
class Bar {
  [x] = foo['y'];
}
```

This is just the practical fallout of supporting ASI in the language unfortunately.

PR #122, opened 2016-09-11
Error on invalid Flow type annotation with default assignment

danharper:
In Flow, a type annotation must come [_before_ a default assignment](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVxhgOZzgE1SgFcA7AYwBcBLOUsADwAoo8AuMAZ0oCdrTsYALxgARAEZRASjABvAL5h0mMACMAhoRIUadRizzCxkjtz4CZCsEA).

```js
// good
function x(foo: string = "1") {} 

// bad
function x(foo = "1": string) {} 
```

Babylon currently parses the right-side type annotation, attaching it to the `AssignmentPattern`.

_I guess this would be a breaking change, though.._

### DIFF
--- a/test/fixtures/flow/type-annotations/105/actual.js
+++ b/test/fixtures/flow/type-annotations/105/actual.js
@@ -1,1 +1,1 @@
-function g(a:number=1, e=1:number) {}
+function g(a:number=1, e:number=1) {}
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -1033,6 +1033,18 @@ export default function (instance) {
     };
   });
 
+  instance.extend("parseMaybeDefault", function (inner) {
+    return function (...args) {
+      const node = inner.apply(this, args);
+
+      if (node.type === "AssignmentPattern" && node.typeAnnotation && node.right.start < node.typeAnnotation.start) {
+        this.raise(node.typeAnnotation.start, "Type annotations must come before default assignments, e.g. instead of `age = 25: number` use `age: number = 25`");
+      }
+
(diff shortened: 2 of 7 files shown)


hzoo:
Yeah I think that was confusing before. I think it's worth making it as a bug fix? @jeffmo

danez:
That is awesome thanks. It is also weird that we had a test to test that the wrong behaviour is working.

Regarding the breaking change, I think we had cases like this in the past where we made the parser more spec compliant which might break peoples workflow if they were relying on this wrong behaviour. This is exactly the same case and imho we should make this a patch, as being compliant to the spec should be important for people no matter if es, flow or jsx. We probably should add a note in the changelog about "more strict spec for ..."

As flow is also failing on this, there are probably also not a lot of cases were people use this (if so it might be a bug in their code anyway) and if people relying on this fact should be 0, at least I can't make up good reason for doing that.

And of course:
https://xkcd.com/1172/

test/fixtures/flow/type-annotations/105/actual.js
```diff
@@ -1 +1 @@
-function g(a:number=1, e=1:number) {}
+function g(a:number=1, e:number=1) {}
```
danez: Weird that we were testing the wrong behaviour is working. 🤔

danharper:
👍  I agree it should be treated as a bug fix. Just thought I'd mention it, as I know in the past I've had at least one file which wasn't being type-checked with Flow, but had some type annotations.

Issue #119, opened 2016-09-10
Improve parsing error message (T6710) 


hzoo:
> Issue originally made by @markelog

Ref https://github.com/babel/babel/issues/3774

### Input code

```js
require("babylon").parse("2++");
```

### Description

Will throw a error - 

"SyntaxError: Assigning to rvalue (1:0)"

Which is pretty cryptic, parsers like esprima, or any other JS engine will show more meaningful message  - "Invalid left-hand side expression in postfix operation"

hzoo:
Moving to https://github.com/babel/babylon/issues/119

markelog:
Moved where?

hzoo:
.. sigh closed this one.

markelog:
Yey!

PR #123, opened 2016-09-12
Rephrase "assigning/binding to rvalue" errors to include context (#119)

motiz88:
This replaces all "Assigning to rvalue" and "Binding rvalue" errors with more detailed variants based on the example in #119.

The implementation is the most straightforward one I could come up with - I add an optional context argument to `checkLVal`, `toAssignable` and `toAssignableList`, a simple string which is interpolated into the final error message (if any).

In some cases I made decisions that make the messages more informative in my view - e.g. making `ArrayPattern` pass through any parent context, so errors look like "invalid left-hand side in assignment expression" rather than "invalid left-hand side in array destructuring pattern".

### Notes

1. The wording is of course not set in stone and I'll happily implement any suggested changes.
2. It's may be preferred aesthetically to move `contextDescription` leftwards in the methods' parameter lists, to avoid all those added `undefined`s. If there's a specific signature you'd like to see here - please let me know.

### DIFF
--- a/src/parser/expression.js
+++ b/src/parser/expression.js
@@ -113,10 +113,10 @@ pp.parseMaybeAssign = function (noIn, refShorthandDefaultPos, afterLeftParse, re
   if (this.state.type.isAssign) {
     let node = this.startNodeAt(startPos, startLoc);
     node.operator = this.state.value;
-    node.left = this.match(tt.eq) ? this.toAssignable(left) : left;
+    node.left = this.match(tt.eq) ? this.toAssignable(left, undefined, "assignment expression") : left;
     refShorthandDefaultPos.start = 0; // reset because shorthand default was used correctly
 
-    this.checkLVal(left);
+    this.checkLVal(left, undefined, undefined, "assignment expression");
 
     if (left.extra && left.extra.parenthesized) {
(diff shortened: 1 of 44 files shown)


hzoo:
I think it looks good overall?

traceur does: `Invalid left-hand side expression in assignment: func()` - dono if it's more helpful to add the left hand side in the error message?

What are the possible left hand values (Identifier/MemberExpression)? - maybe we can suggest that

motiz88:
Working on that merge conflict now.

motiz88:
I'm personally not crazy about the idea of adding bits of code to error messages, for a number of reasons (though it's probably a matter of taste more than anything).

* We wouldn't always want to clutter the message with the full left-hand code (imagine the output for `func ( /* lots of arguments and complex expressions here... */ ) = 5`), so we might introduce formatting and abbreviating logic, which can be nontrivial and lead to bloat.

* There's already location information in the error object - my ideal workflow would be one where a tool / frontend just consumes and visualizes _that_ (along with a clear, concise message), without babylon trying to hint at it separately via the error message.

hzoo:
> We wouldn't always want to clutter the message

fair enough

Issue #113, opened 2016-09-08
Better error output for await in non-async functions


danez:
> Issue originally made by @SystemParadox on babel/babel#2387

``` js
function test() {
    await foo();
}
```

Throws: `Unexpected token(2:8)`

This is a bit misleading. Can we change this to say something like "await can only be used in async functions"?

Thanks.

motiz88:
Hi! I'm trying to implement this and learn some Babylon internals at the same time.

**tl;dr: I have a question about a specific implementation idea - see the very end**

### :thought_balloon: Larger discussion: `await` spec

The spec says [here](https://tc39.github.io/ecmascript-asyncawait/#async-function-definitions) under note 1 that (emphasis mine)

> `yield` is _Identifier_ or _YieldExpression_ [ ... ] depending on context. **`await` is always parsed as an _AwaitExpression_ but is a Syntax Error via static semantics.** [ ... ]

Does this mean we can skip trying to disambiguate `await` and just treat it as starting an AwaitExpression everywhere (i.e. immediately an error if not in an async context)? Probably not. I must be missing something. But it would make fixing this issue quite trivial. Moving on...

### :point_right: The issue at hand

I'm assuming the intent here is to optimistically parse `await` as Identifier in non-async contexts, the way it's already being done. But then if the parse turns out to be a dead end, we should "remember" this choice and emit a descriptive error on the `await` - not at the dead end.

I came up with the following logic:

1. When an Identifier with `name === "await"` is parsed, if we didn't enter `parseAwait` because we're in a non-async context, save the identifier's position in `state.potentialIllegalAwaitAt` and keep parsing normally.
2. When parsing a node fails via `unexpected()`, and if `state.potentialIllegalAwaitAt >= node.start`, raise an "await can only be used in async function" error instead of the default "unexpected token".

I'm not 100% sure how to intercept the `unexpected()` occurrences elegantly while still having access to the current node - which is needed for the check in (2) above. My first approach involved patching a _lot_ of node types and looked a bit like this:

<details>
```javascript
pp.checkIllegalAwaitAndRethrow = function (err, node) {
  if (this.state.potentialIllegalAwaitAt >= node.start) {
    this.raise(this.state.potentialIllegalAwaitAt, "await can only be used in async functions");
  } else {
    throw err;
  }
};

pp.semicolonCheckIllegalAwait = function (node) {
  try {
    this.semicolon();
  } catch (err) {
    this.checkIllegalAwaitAndRethrow(err, node);
  }
[...]

PR #125, opened 2016-09-13
[WIP] Improve error message for await-outside-async

motiz88:
This is an attempt to implement #113. I would love some early feedback/pointers, as well as help completing the following list which reflects the current state of this branch. I only added tests for true positives and true negatives, so they should all pass despite the work being incomplete.
### :sunglasses: True positives :thumbsup:

The following all raise `await can only be used in async functions` correctly (to my understanding).
- `await promise`
- `return await promise`
- `var x = await promise`
- `x[await promise]`
- `x(await promise)`
- `x(await promise, 0)`
- `await await`
- `await yield`
- `await this`
- `await super`
- `await {}`
- `function* a() {yield await 0}`
- `await new Promise`
- `await.0`
### :sunglasses: True negatives :thumbsdown:

The following all raise `Unexpected token` etc, correctly (to my understanding).
- `await}`
- `await(` - parseExprSubscripts
- `await.`
- `await[`
- `await+`
- `await``
- `await/`
- `await*`
- `await[0;`
- `await(0`
- `await instanceof`
### :disappointed: False positives :thumbsup:

The following all raise `await can only be used in async functions` _incorrectly_ (only in actual non-async contexts, though).
##### parseExpressionStatement
- `await var`
- `await)`
- `await~`
- `await{`
- `await x(`
- `await x((`
##### parseSubscripts
- `x[await`
##### parseParenAndDistinguishExpression
- `(await`
### :disappointed: False negatives :thumbsdown:

Cases which should raise `await can only be used in async functions` but raise some other error: None that I know of.
### :sunglasses: No error at all :ok_hand:
- `await ```


hzoo:
Should we change the wording to say something like: `can only be used in an immediate async function` (or just to say the parent function should be async) - we could give an example too

danez:

Thanks for your effort.

PR #127, opened 2016-09-15
Add tests for export extensions

danez:
The case which includes a namespaced and default import was not tested yet

This should increase code coverage for this case.


jdalton:
I noticed that this will error:
```js
export def4, * as ns4, { c as e } from "../misc/abc";
```
The bit that's throwing it off is the last specifier `, { c as e }`.

jdalton:
Related to https://github.com/benjamn/reify/pull/162.

PR #135, opened 2016-09-20
Fix `static` property falling through in `declare class` Flow AST

danharper:
Given the following:

```js
declare class X {
	a: number;
	static b: number;
	c: number;
}
```

`c` would get marked as `static` as it's not being reset between properties.

https://astexplorer.net/#/4Bi9Ce53LA


danez:
Awesome thanks

PR #137, opened 2016-09-22
Fix duplicate check

danez:
Fix comment in #107

@kaicataldo Can you check if this is correct?


src/tokenizer/state.js
```diff
@@ -123,7 +123,7 @@ export default class State {
 
   // Names of exports store. `default` is stored as a name for both
   // `export default foo;` and `export { foo as default };`.
-  exportedIdentifiers: {[id:string]: boolean};
+  exportedIdentifiers: Array<string;
```
hzoo: Array<string>;

danez: 👍

hzoo:
cc @dlmr

dlmr:
Looks good! LGTM 👍

src/tokenizer/state.js
```diff
@@ -43,7 +43,7 @@ export default class State {
     this.containsEsc = this.containsOctal = false;
     this.octalPosition = null;
 
-    this.exportedIdentifiers = {};
+    this.exportedIdentifiers = [];
```
STRML: Any reason to not use Object.create(null); here instead?

PR #140, opened 2016-09-23
Don't do the duplicate check if not an identifier

hzoo:
Passes locally

TODO: should add logic to handle it correctly in a future PR when we have time


loganfsmyth:
LGTM pending tests.

Issue #112, opened 2016-09-08
Destructing function parameters with "set" property in it


danez:
> Reported by @pavelkornev

Following snippet:

```
function A({set}) {
  set = "a"
}
```

compiles to:
```
"use strict";

function A(_ref) {
  var set = _ref.set;

  set = "foo";
}
```

https://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=function%20A(%7Bset%7D)%20%7B%0A%20%20set%20%3D%20%22a%22%0A%7D

But when we try to assign default value to "set":
```
function A({set="bar"}) {
  set="foo"
}
```

it crashes with the following error:
```
repl: Unexpected token (1:15)
> 1 | function A({set="bar"}) {
    |                ^
  2 |   set="foo"
  3 | }
```
https://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=function%20A(%7Bset%3D%22bar%22%7D)%20%7B%0A%20%20set%3D%22foo%22%0A%7D

On other hand if we assign alias to it:
```
function A({set:set_="bar"}) {
  set="foo"
}
```

It compiles just fine:
```
"use strict";

function A(_ref) {
  var _ref$set = _ref.set;
  var set_ = _ref$set === undefined ? "bar" : _ref$set;

  set = "foo";
}
```
https://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&code=function%20A(%7Bset%7D)%20%7B%0A%20%20set%3D%22foo%22%0A%7D

If "set" reserved why does it work in first case? If it's not why second case doesn't work? I consider this as a bug of compiler, am i right?

PR #142, opened 2016-09-24
get / set are valid property names in default assignment

jezell:
```
function something({ set = null }) {
}
```

this is valid syntax, but currently fails to compile with babel

### DIFF
--- a/src/parser/expression.js
+++ b/src/parser/expression.js
@@ -774,8 +774,8 @@ pp.parseObjPropValue = function (prop, startPos, startLoc, isGenerator, isAsync,
     return this.finishNode(prop, "ObjectProperty");
   }
 
-  if (!prop.computed && prop.key.type === "Identifier" && (prop.key.name === "get" || prop.key.name === "set") && (!this.match(tt.comma) && !this.match(tt.braceR))) {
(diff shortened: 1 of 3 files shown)


danez:
Thank you, this was also reported in #112.

src/parser/expression.js
```diff
@@ -774,7 +774,7 @@ pp.parseObjPropValue = function (prop, startPos, startLoc, isGenerator, isAsync,
     return this.finishNode(prop, "ObjectProperty");
   }
 
-  if (!prop.computed && prop.key.type === "Identifier" && (prop.key.name === "get" || prop.key.name === "set") && (!this.match(tt.comma) && !this.match(tt.braceR))) {
+  if (!isPattern && !prop.computed && prop.key.type === "Identifier" && (prop.key.name === "get" || prop.key.name === "set") && (!this.match(tt.comma) && !this.match(tt.braceR))) {
     if (isGenerator || isAsync || isPattern) this.unexpected();
```
danez: `|| isPattern` can now be removed as it will never be true.

jezell: good catch, fixed.

danez:
Thank you

danez:
Upps wrong button :D

Issue #139, opened 2016-09-23
Breaking change: eslint-plugin-import broken 6.10.0->6.11.1


scottnonnenberg:
Good build (6.10.0): https://circleci.com/gh/scottnonnenberg/blog/25
Bad build (6.11.1): https://circleci.com/gh/scottnonnenberg/blog/26

Seems that handling of named `export` statements has regressed:

```
SyntaxError: /home/ubuntu/blog/src/util/typography.js: `undefined` has already been exported. Exported identifiers must be unique. (22:13)
  20 | 
  21 | export const { rhythm } = typography;
> 22 | export const { TypographyStyle } = typography;
     |              ^
  23 | export const { fontSizeToMS } = typography;
```

<details>
Here's the whole set of the dependency changes between these two builds:

```
56c56
< │ │ ├─┬ optionator@0.8.1
---
> │ │ ├─┬ optionator@0.8.2
58c58
< │ │ │ ├── fast-levenshtein@1.1.4
---
> │ │ │ ├── fast-levenshtein@2.0.4
68c68
< │ │ │ ├─┬ ajv@4.7.2
---
> │ │ │ ├─┬ ajv@4.7.3
70c70
< │ │ │ ├── ajv-keywords@1.0.0
---
> │ │ │ ├── ajv-keywords@1.1.0
234c234
< │ ├── babylon@6.10.0
---
> │ ├── babylon@6.11.1
472,473c472,473
< │ │ │ ├─┬ global@4.3.0
< │ │ │ │ ├─┬ min-document@2.18.1
---
> │ │ │ ├─┬ global@4.3.1
> │ │ │ │ ├─┬ min-document@2.19.0
614c614
< │ ├── highlight.js@9.6.0
---
> │ ├── highlight.js@9.7.0
814c814
< │ │ │ ├── base64-js@1.1.2
---
> │ │ │ ├── base64-js@1.2.0
843c843
< │ │ │ ├── pbkdf2@3.0.7
---
> │ │ │ ├── pbkdf2@3.0.8
922c922
< │ │ │ ├── caniuse-db@1.0.30000534
---
> │ │ │ ├── caniuse-db@1.0.30000539
1054c1054
< │ │ │   │ ├── rsvp@3.3.1
---
> │ │ │   │ ├── rsvp@3.3.2
```
</details>

hzoo:
Ok it doesn't account for destructuring @kaicataldo @danez

kaicataldo:
[...]

PR #144, opened 2016-09-24
Check for duplicate named exports in exported destructuring assignments

kaicataldo:
This is a follow-up to https://github.com/babel/babylon/issues/139. I did a little refactoring - passing `isDefault` all the way through seems unnecessary, in retrospect. Thoughts on this approach? Any cases I missed?


src/parser/statement.js
```diff
@@ -940,6 +938,22 @@ pp.checkExport = function (node, checkNames, isDefault) {
   }
 };
 
+pp.checkDeclaration = function(node, isDefault) {
```
kaicataldo: We actually might not need to pass through `isDefault` here, since `default` is a reserved word and can't be the name for an `Identifier`

test/fixtures/es2015/modules/duplicate-named-export-destructuring/actual.js
```diff
@@ -1,2 +1,11 @@
-export const { rhythm } = typography;
-export const { TypographyStyle } = typography;
+export const { foo } = bar;
```
kaicataldo: These are all unique identifiers and should parse without error

src/parser/statement.js
```diff
@@ -909,23 +909,21 @@ pp.checkExport = function (node, checkNames, isDefault) {
     // Check for duplicate exports
     if (isDefault) {
       // Default exports
-      this.checkDuplicateExports(node, "default", isDefault);
+      this.checkDuplicateExports(node, "default");
     } else if (node.specifiers && node.specifiers.length) {
       // Named exports
       for (let specifier of node.specifiers) {
         const name = specifier.exported.name;
         if (name === "default") isDefault = true;
```
loganfsmyth: Looks like this isn't needed.

src/parser/statement.js
```diff
@@ -940,15 +936,31 @@ pp.checkExport = function (node, checkNames, isDefault) {
...
+  if (node.type === "ObjectPattern") {
+    for (let prop of node.properties) {
+      this.checkDeclaration(prop);
+    }
+  } else if (node.type === "ArrayPattern") {
+    for (let elem of node.elements) {
+      this.checkDeclaration(elem);
+    }
+  } else if (node.type === "ObjectProperty") {
+      this.checkDeclaration(node.value);
```
danez: to much indentation

test/fixtures/es2015/modules/duplicate-named-export-destructuring/actual.js
hzoo: Cool was wondering about a passing test 👍

kaicataldo:
Updated - thanks!

kaicataldo:
@danez I know we spoke in Slack, but wasn't sure if there is anything I need to do for the failing codecov checks?

PR #149, opened 2016-09-29
Fix parsing object rest spread

### LINKED ISSUE #148: Fix destruction parsing
Ref https://github.com/babel/babel/issues/4222

## Description

The spec in https://github.com/sebmarkbage/ecmascript-rest-spread/blob/master/Rest.md defines

let { ...x, y, z } = obj; // syntax error
let { x, ...y, ...z } = obj; // syntax error
Both of which appear to be allowed without issue.

Source: http://stackoverflow.com/questions/36429161/should-rest-property-of-object-binding-pattern-be-the-last

danez:
This makes object-rest-spread behave according to spec and only
allow one spread/rest operator and enforces it to be the last
param in the object.

Also move all object-rest-spread tests to a own folder.

The fix is pretty simple, as soon as we parsed the spread operator we expect the object to end.

The only thing I'm not 100% sure is if trailing commas are aloud, but I would assume it is the same as with array spread to be not allowed. (see test 8) //cc @sebmarkbage

Fixes #148


test/fixtures/flow/type-parameter-declaration/class-with-jsx/expected.json
```diff
@@ -0,0 +1,33 @@
+{
```
hzoo: This file looks unrelated?

src/parser/expression.js
```diff
@@ -713,7 +713,8 @@ pp.parseObj = function (isPattern, refShorthandDefaultPos) {
       prop = this.parseSpread();
       prop.type = isPattern ? "RestProperty" : "SpreadProperty";
       node.properties.push(prop);
-      continue;
+      this.expect(tt.braceR);
```
hzoo: Is it easy to have have more specific error messages for these?

`let { ...x, y, z } = obj;` - spread has to be the last element (would result in the same value as the right hand side)
`let { x, ...y, ...z } = obj;` - cannot have multiple spreads (would result in the same value for both y and z)?

sebmarkbage:
The single __rest__ at the end only applies to binding `let { x, ...y } = obj;` and assignment `({ x, ...y } = obj)`. But is doesn't apply to __spread__ `let obj = { ...x, y, ...z }; // works`.

I think you need to make this conditional on the `isPattern` flag, no?

hzoo:
Yeah only for rest 👍 

The failing test from babel is example of this (valid)

```
babel-plugin-transform-object-rest-spread/object rest spread expression:
     SyntaxError: /home/travis/build/babel/babylon/build/babel/packages/babel-plugin-transform-object-rest-spread/test/fixtures/object-rest-spread/expression/actual.js: Unexpected token (1:10)
  > 1 | ({ x, ...y, a, ...b, c });
```

danez:
Is trailing comma allowed in this case?

```
const { x, ...s, } = obj;
```

test/fixtures/flow/type-parameter-declaration/class-with-jsx/expected.json
danez: Upps removed

sebmarkbage:
@danez No, trailing comma is not allowed.

hzoo:
:+1: Other than I think we could give a better message

src/parser/expression.js
danez: Done, I also added a separate message when having an invalid trailing comma

src/parser/expression.js
```diff
@@ -714,8 +716,17 @@ pp.parseObj = function (isPattern, refShorthandDefaultPos) {
       prop.type = isPattern ? "RestProperty" : "SpreadProperty";
       node.properties.push(prop);
       if (isPattern) {
-        this.expect(tt.braceR);
-        break;
+        const position = this.state.start;
+        if (firstRestLocation !== null) {
+          this.unexpected(firstRestLocation, "Cannot have multiple rest elements in destruction expression");
```
hzoo: Maybe `"Cannot have multiple rest elements in a destructuring assignment"` or `"Cannot have multiple rest elements when destructuring"`?

danez: I did not use `assignment` as I think this is not an assignment if I'm not mistaken:

```
function ({a,b,...c,...c}) {}
```
But this would issue the same error. Should add a test for it if there isn't one yet.

Can use `"Cannot have multiple rest elements when destructuring"`.

hzoo: > Can use "Cannot have multiple rest elements when destructuring". 👍

src/parser/expression.js
```diff
@@ -753,6 +769,10 @@ pp.parseObj = function (isPattern, refShorthandDefaultPos) {
     node.properties.push(prop);
   }
 
+  if (firstRestLocation !== null) {
```
hzoo: These can be moved right below the `while (!this.eat(tt.braceR)) {` block right? - I guess it was left at the end as convention? Works either way

danez: no it has to be at the end after the loop, because otherwise the multiple destructuring check does not work.

Basically the logic is:
```
loop through all params:
   if spread
       if havent seen spread before AND this is not the end of the object pattern: record that we have seen a spread
       else if havent seen spread before: warning that this is already the second spread

if we finished the loop and recorded a spread: error that it was not at the end
```

hzoo: Oh I think I misread the last `}`!

hzoo:
Just gotta fix the error messages in the tests 👍

istarkov:
This PR does not work for arrow functions,

```javascript
const x = ({ ...props, y }) => 1; // THIS WORK
function xx({ ...props, y }) { return 1;} // THIS FAIL
```

danez:
Good catch, thanks for reporting.

istarkov:
This catch will cost me a day to fix hundreds of my code lines ;-) (Will read the specs in the future)

PR #150, opened 2016-09-29
[WIP] Slightly better error messages

danez:
Instead of `Unexpected token` we now print `Unexpected token, expected <type.label>` in the cases where we have the information. (see test failures for examples)

I did not yet fix the tests, wanted to hear first what you think about that. It is not that big of an improvement but maybe better than what was there before.


danharper:
👍  I'd like to see this.

hzoo:
Got a timeout error and one of the dependencies is failing node 0.12 for some reason but we can still merge

PR #161, opened 2016-10-07
Property variance type annotations for Flow plugin

samwgoldman:
Non-method properties and indexers of object types, declare class, and
interfaces can be "positive" or "negative." Class fields, but again not
methods, can also have variance.

This PR generalizes the variance annotations for type parameters into a
new node type, and reuses that node for those properties.

The code for object types is reused for interfaces and declare classes.
The changes there are straightfoward.

The code for class fields is reused for object literals, which do not
support variance annotations (currently). This code is a bit sketchy,
because we always parse variance annotations in the `parsePropertyName`
extension, then error in a the subsequent parse phase for object
literals (`parseObjPropValue`) or class methods (`parseClassMethod`).


samwgoldman:
Tests fail due to missing babel support, which is added in the referenced PR to Babel.

Is it possible to bump these versions in lock step, or do I need to break up this change into multiple steps?

src/plugins/flow.js
```diff
@@ -726,6 +730,21 @@ pp.typeCastToParameter = function (node) {
...
+pp.flowParseVariance = function() {
+  let variance = null;
+  if (this.match(tt.plusMin)) {
+    variance = this.startNode();
+    if (this.state.value === "+") {
+      variance.kind = "plus";
+    } else if (this.state.value === "-") {
+      variance.kind = "minus";
+    }
+    this.eat(tt.plusMin);
```
danez: We could simplify that here and do `this.eat(tt.plusMin)` already in the `if ()` and remove this line then.

src/plugins/flow.js
```diff
@@ -380,6 +379,10 @@ pp.flowParseObjectType = function (allowStatic, allowExact) {
       }
       if (this.isRelational("<") || this.match(tt.parenL)) {
         // This is a method property
+        if (variance) {
+          this.unexpected(variance.start);
+          variance = null;
```
danez: This doesn't makes sense as `this.unexpected()` throws anyway. Maybe we want to do `delete ...` here like in the other places?

samwgoldman: Good catch! These two assignments are totally bogus :P

src/plugins/flow.js
samwgoldman: Will change the `eat(plusMin)` to `next()`, which will hopefully clarify the intent better. The tests do exercise the node positioning, so regression is unlikely here.

danez: Ah I see, sorry.

samwgoldman:
Heads up, I'm going to change this to roll back the addition of a new variance node. Instead, I will continue to use the `"plus" | "minus" | null` format that currently exists for type parameters.

samwgoldman:
OK, I removed the new node type, so this is no longer a breaking change. In order to preserve the existing error messages, I needed to pass around the position of the variance annotation out-of-band, which is a bit hacky, because we need to be careful to clean up after ourselves with `delete`.

In the future, when Babel approaches a major version and we can make the breaking change, I'd like to go back to having a separate node type for variance annotations.

src/plugins/flow.js
```diff
@@ -1052,9 +1055,10 @@ export default function (instance) {
   instance.extend("parseObjPropValue", function (inner) {
     return function (prop) {
       if (prop.variance) {
-        this.unexpected(prop.variance.start);
+        this.unexpected(prop.start);
```
danez: Should this be `prop.variancePos`

samwgoldman: 👍 Yup. In this case they are one-in-the-same, but this should definitely be `variancePos` for clarity. We wouldn't need `variancePos` except for static properties on classes, which share the same AST building code.

PR #163, opened 2016-10-08
[dynamic-import] Implementing import() syntax

kesne:
There's a new stage-2 proposal that adds a new special `import()` function. This implements the syntax behind a `importFunctions` plugin.

Here's the proposal: https://github.com/domenic/proposal-import-function

cc: @domenic


domenic:
Thanks so much for working on this; this'll be very helpful.

As a nit, I think it might be better to name it importFunctionLike so people don't get the wrong idea. It's more like super() than it is like a function. I'll probably rename the proposal repo similarly.

kesne:
Yeah I thought about that working on this, that's probably not a bad idea. I'm also not :100: on `ImportCallExpression` in the AST.

src/parser/expression.js
```diff
@@ -300,7 +300,8 @@ pp.parseSubscripts = function (base, startPos, startLoc, noCalls) {
       let node = this.startNodeAt(startPos, startLoc);
       node.callee = base;
       node.arguments = this.parseCallExpressionArguments(tt.parenR, possibleAsync);
-      base = this.finishNode(node, "CallExpression");
+      base = this.finishNode(node, this.state.inImportCall ? "ImportCallExpression" : "CallExpression");
```
kesne: @loganfsmyth Alright I updated the code to do that. Ended up being a pretty clean solution! Thanks for the recommendation.

src/parser/expression.js
```diff
@@ -394,6 +404,7 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
       if (this.state.inGenerator) this.unexpected();
 
     case tt.name:
+      if (!this.hasPlugin("importFunctions") && this.state.type === tt._import) this.unexpected();
```
loganfsmyth: Looks like this isn't needed now?

kesne: Yep, good catch! Fixed.

src/parser/statement.js
```diff
@@ -98,6 +98,18 @@ pp.parseStatement = function (declaration, topLevel) {
...
+      if (this.hasPlugin("importFunctions")) {
+        // peek ahead and see if next token is a starting paren
+        let state = this.state.clone();
+        this.next();
+        if (this.match(tt.parenL)) {
+          this.state = state;
+          break;
+        } else {
+          this.state = state;
+        }
```
loganfsmyth: I _think_ this can be simplified to this, though I'm not an expert on Babylon.

    if (this.hasPlugin("importFunctions") && this.lookahead().type === tt.parenL) break;

ljharb:
Is it a good idea to ensure in the tests that it parses in both sloppy and strict mode?

kesne:
@ljharb Not a bad idea. I can also add a test case for calling in module mode (right now I just run in script mode and assume that if it passes in script, it passes in module).

src/parser/expression.js
```diff
@@ -385,6 +385,16 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
       }
       return this.finishNode(node, "Super");
 
+    case tt._import:
+      if (!this.hasPlugin("importFunctions")) this.unexpected();
+
+      node = this.startNode();
+      this.next();
+      if (!this.match(tt.parenL)) {
```
kesne: It looks like this case can never be hit, because the parser runs through statements first, and would throw unexpected trying to parse the import function.

Should I remove it knowing that, or keep it for safety?

loganfsmyth: I wondered that too, good question, curious to see what others think.

danez: Also codecoverage says we never hit this block, but as we do the same for super maybe be consistent and leave it here?

kesne: I figured out that this can be hit if you do something that parses as an expression first. For example, inside of a function `return import.fails()` hits this case.

I've added a test for this case to prevent any regressions and to bump the code coverage.

kesne:
@domenic Considering this: https://github.com/domenic/proposal-import-function/issues/15 do you think that we should enforce the single argument syntax in the parser now?

domenic:
Yeah, that's probably a good idea... not sure what the spec grammar will look like yet, which might impact your AST...

src/parser/expression.js
```diff
@@ -385,6 +385,16 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
       }
       return this.finishNode(node, "Super");
 
+    case tt._import:
+      if (!this.hasPlugin("importFunctions")) this.unexpected();
```
hzoo: Doesn't have to be done in this PR since it's a separate issue but we should make a better error message here (when we have a `this.hasPlugin(`)

kesne: I agree that would be nice. I did this because it seems like this is how the rest of the plugins handle it.

kesne:
Alright, I added a check to the arguments which fails parse on more than one argument. The only remaining line that I don't have code coverage for is the check for the plugin in expressions. If you folks think that'd be good to have I can do that.

src/parser/expression.js
hzoo: Yeah it's totally fine just a note so I don't forget - I can just make an actual issue

domenic:
Updates from the spec side:

- The new alphanumeric name of the import() proposal is "proposal-dynamic-import"
- The spec now requires exactly one argument. (Might be worth testing the no-argument `import()` case too!)

Exciting!

kesne:
Alright, I renamed the plugin `dynamicImport`, and updated it to require exactly one argument and added a test to check for the no-arg case!

kesne:
Is there anything else that I can do to help get this out?

hzoo:
Kind of trivial but maybe a test that `import()` fails to parse without the plugin turned on?

src/parser/expression.js
```diff
@@ -385,6 +388,16 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
       }
       return this.finishNode(node, "Super");
 
+    case tt._import:
+      if (!this.hasPlugin("dynamicImport")) this.unexpected();
+
+      node = this.startNode();
+      this.next();
+      if (!this.match(tt.parenL)) {
+        this.unexpected();
```
hzoo: Maybe we can do a 
```
this.unexpected(null, `Unexpected token, expected (`);
```

https://github.com/babel/babylon/pull/150/files#diff-51a873a9697282bed0c15722a6ccf4e6R77

not sure if you can just use this.expect instead

hzoo: Actually we can use https://github.com/babel/babylon/pull/172

to do `this.unexpected(null, tt.parenL);`

hzoo: yeah

PR #165, opened 2016-10-09
Fix: Allow identifier `async` for default param in arrow expression

### LINKED ISSUE #118: Parse error when using "async" as a parameter name.
> Reported by @juandopazo

When using async as a parameter name, we get an "Unexpected token" error.

See http://babeljs.io/repl/#?evaluate=false&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=const%20test%20%3D%20(%7Basync%20%3D%20true%7D)%20%3D%3E%20%7B%0A%E2%80%82%E2%80%82console.log(async)%3B%0A%7D%3B

```js
const test = ({async = true}) => {
  console.log(async);
};
```

kaicataldo:
fixes #118


danez:
Thank you

PR #170, opened 2016-10-12
Fix crash when exporting with destructuring and sparse array

jfmengels:
Hi there!

The tests for eslint-plugin-import are currently [failing](https://travis-ci.org/benmosher/eslint-plugin-import/jobs/167153714), with the `Cannot read property 'type' of null` error.

I tracked down the error to Babylon and the new [checks to whether a duplicate export exists](https://github.com/babel/babylon/blob/master/src/parser/statement.js#L945-L947), and have fixed it in this PR.

I have created a minimal viable example, which is available in my first commit, to show that the tests fail unexpectedly with the error message above. It looks like this
```js
export const { foo: [ ,, qux7 ] } = bar;
```

(Actually, `export const [,, a] = {}` would have sufficed, but I just noticed it, and have already updated the `expected.json` file in the tests, which took me a while :sweat_smile:... let me know if you prefer I use this form instead, or add both cases in the tests)

The problem is that the code [here](https://github.com/babel/babylon/blob/master/src/parser/statement.js#L945-L947) loops through the elements inside the destructuring expression, but forgets to filter out the `null` ones, that you get when you have a sparse array like `[,,a]`, before directly accessing the `type` property.

I added a null check around the recursive call, and that fixes the problem.

Let me know if you need more info, more work or anything else :)


hzoo:
Nope that looks good/makes sense! Thanks a lot!!

hzoo:
>  and have already updated the expected.json file in the tests, which took me a while

@jfmengels If you create an actual.js without an expected it should automatically make the file - hopefully you didn't mean you did it manually 😅 , but if so next time!

Will merge + release after travis is done

jfmengels:
> hopefully you didn't mean you did it manually :sweat_smile: , but if so next time!

Haha, I did... :sweat_smile:

I didn't know how tests were created, so I just updated an existing test, which already had the file. Deleting it was not the most intuitive thing to do :smile:.

Honestly I just thought you were working like a madman to have proper `expected.json` files. This makes much more sense :smile:. (Though, if you don't write it at least in part manually, I'm not sure what exacly you're testing, but that's another topic). I myself cheated a bit by using astexplorer.net, and then manually fixing every error the tests showed me. I will definitely remember it next time.

> Will merge + release after travis is done

Awesome! Thanks for being so reactive!

PR #172, opened 2016-10-14
Annotate more errors with expected token

motiz88:
Expanding on #150, this allows `unexpected()` to accept the expected token type instead of a message string. This overload is then used in a couple more places (that essentially reimplement `expect()`) to construct an `Unexpected token, expected FOO` message.

I can imagine a future enhancement to `unexpected()` in which it would similarly take an array of token types; then we could say `expected FOO or BAR or BAZ` where it makes sense.


src/parser/util.js
```diff
@@ -67,18 +67,22 @@ pp.isLineTerminator = function () {
 // pretend that there is a semicolon at this position.
 
 pp.semicolon = function () {
-  if (!this.isLineTerminator()) this.unexpected();
+  if (!this.isLineTerminator()) this.unexpected(null, tt.semi);
```
danez: Also a newline would be allowed here I think.

danez:
@motiz88 Can you have a look and see why the babel tests now fail with this PR? One sems fine, but have no idea about the other failure.

motiz88:
@danez Those are both assertions on messages that have changed in this PR. Not sure why one of them (`transformation/errors/syntax`) doesn't _say_ so in the log and instead just throws the SyntaxError. Will prepare a Babel PR to go along with this one.

PR #175, opened 2016-10-14
Increase test coverage

motiz88:
I intend to do a few of these - here's the first batch of new test cases written explicitly to hit previously uncovered lines/branches. I've also added one `// istanbul ignore next` and removed one line of dead code.

EDIT: I've added a few more in the same vein. Anyone who wants to review and merge - tell me and I'll stop here and continue on another branch.


src/parser/statement.js
```diff
@@ -668,7 +668,6 @@ pp.parseClassBody = function (node) {
 
     method.static = isMaybeStatic && !this.match(tt.parenL);
     if (method.static) {
-      if (isGenerator) this.unexpected();
```
danez: This was never triggered?

motiz88: Yeah, `method.static` hinges on `this.match(tt.name)` and `isGenerator` (subsequently) on `this.eat(tt.star)`. So they are mutually exclusive.

danez:
Holy crap, I wasn't aware we have so much dead code. I'm going to merge this now.
Awesome work.

Issue #169, opened 2016-10-12
Better error message on `unexpected token` when it's because a plugin isn't turned on


hzoo:
Ref https://github.com/babel/babylon/pull/163#discussion_r82658832

Example `if (!this.hasPlugin("dynamicImport")) this.unexpected();` or

```
Module build failed: SyntaxError: Unexpected token (1:6)

> 1 | import();
    |       ^
  2 |
```

If a user is using an experimental syntax and we error then we should tell them a message so it can be fixed.

We don't necessarily know they are using babel/babylon though so the error message can give some hints:
- ask the user to turn on the corresponding plugin in babylon
- ask the user to add the syntax/transform plugin in babel
- ask the user to use `parserOpts: { plugins: ['syntax'] }`

motiz88:
Going to tackle this. I'm thinking of an `expectPlugin` helper method to replace `if (!hasPlugin(...)) unexpected()`, and maybe a small map of known Babylon plugins to Babel plugins, to support helpful messages.

nathanhammond:
In-progress PR: https://github.com/babel/babylon/pull/178

PR #178, opened 2016-10-14
Point out when a plugin needs to be enabled for encountered syntax (#169)

motiz88:
Ref #169
This mainly replaces the `if (!this.hasPlugin("foo")) this.unexpected();` idiom with a new parser helper method, `expectPlugin`.

Right now the error is worded like this: `This experimental syntax requires the foo parser plugin. Check out babel-plugin-syntax-foo, babel-plugin-transform-foo or the parserOpts.foo Babel option.` Open for different suggestions of course.

Another idiom we may want to change is where we currently check for things like `this.hasPlugin("asyncGenerators") && this.eat(tt.star)`. Does it make sense to flip the order in some places? Match the special token (here `*`) without advancing, then assert that we have the plugin, then advance. In cases where we can be sure of the user's intent this looks like a usability win.


motiz88:
Really not sure why eslint is saying that `node` is unused [here](https://github.com/motiz88/babylon/blob/df7868c9a369ff0d7a0d45c7a4174bc433c405ce/src/parser/expression.js#L376) 😕

motiz88:
#182 will probably take care of most of the gap in code coverage

motiz88:
@danez Do you have an opinion on `this.hasPlugin("asyncGenerators") && this.eat(tt.star)` ➡️ `this.match(tt.star) && this.expectPlugin("asyncGenerators"), this.next()`?

danez:
I think this would also be the only way to do it in this cases right? Because otherwise by doing `this.eat() && this.expectPlugin()`would change the location of the error.

motiz88:
Yeah. The thing to watch out for, if we eagerly `match() && expectPlugin()`, is cases where there is an alternative correct parse we can't rule out yet. Async generators are not generally such a case - after `async function`, if we see `*` it can't be anything but an error, so it's a safe bet the user meant it as an async generator.

danez:
True, so probably we need to decide for each occurrence separately.

motiz88:
OK, I'll just add some of the easy cases here real quick.

src/parser/expression.js
```diff
@@ -436,18 +436,17 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
...
-        this.next();
-        let oldInFunction = this.state.inFunction;
-        let oldLabels = this.state.labels;
-        this.state.labels = [];
-        this.state.inFunction = false;
-        node.body = this.parseBlock(false, true);
-        this.state.inFunction = oldInFunction;
-        this.state.labels = oldLabels;
-        return this.finishNode(node, "DoExpression");
-      }
```
danez: We had a fall-through here before without `doExpression`, but that looks like a mistake anyway. Why should do fall-through to regexp?

src/parser/expression.js
```diff
@@ -531,10 +530,20 @@ pp.parseFunctionExpression = function () {
 
 pp.parseMetaProperty = function (node, meta, propertyName) {
   node.meta = meta;
+
+  if (meta.name === "function" && propertyName === "sent") {
+    if (this.isContextual(propertyName)) {
+      this.expectPlugin("functionSent");
+    } else if (!this.hasPlugin("functionSent")) {
+      // They didn't actually say `function.sent`, just `function.`, so a simple error would be less confusing.
+      this.unexpected();
```
danez: Uncovered. We could test this I think easily.

src/parser/statement.js
```diff
@@ -228,9 +226,10 @@ pp.parseForStatement = function (node) {
   this.state.labels.push(loopLabel);
 
   let forAwait = false;
-  if (this.hasPlugin("asyncGenerators") && this.state.inAsync && this.isContextual("await")) {
-      forAwait = true;
-      this.next();
+  if (this.isContextual("await") && this.state.inAsync) {
```
danez: Maybe switch the order here like it was before, checking the bool value first is maybe also faster in case it is false?
Or did you have something in mind by changing the order.

src/parser/statement.js
```diff
@@ -688,7 +686,11 @@ pp.parseClassBody = function (node) {
 
     let isAsyncMethod = !this.match(tt.parenL) && !method.computed && method.key.type === "Identifier" && method.key.name === "async";
     if (isAsyncMethod) {
-      if (this.hasPlugin("asyncGenerators") && this.eat(tt.star)) isGenerator = true;
+      if (this.match(tt.star)) {
+        this.expectPlugin("asyncGenerators");
+        this.next();
```
danez: We have this pattern quite often `if (match) expectPlugin(), next()` So maybe we could make this easier in future by doing:
```
if (this.expectPluginIfMatch("plugin", tt.start)) {
 is = true
}
```
What do you think?
But should do that anyway later, to not make this PR even more bigger.

src/parser/util.js
```diff
@@ -86,3 +87,25 @@ pp.unexpected = function (pos, messageOrType = "Unexpected token") {
...
+
+pp.expectPlugin = function (name) {
+  if (!this.hasPlugin(name)) {
+    const pluginInfo = plugins[name];
+    const pointers = [];
+    if (pluginInfo) {
+      if (pluginInfo.babelSyntaxPlugin) {
+        pointers.push(`babel-plugin-${pluginInfo.babelSyntaxPlugin}`);
+      }
+      if (pluginInfo.babelTransformPlugins) {
```
danez: Should we make `babelTransformPlugins` mandatory to be an array, so we don't have to check here?
And also as long as we have syntax-plugins for everything, maybe also remove the if for it?

src/plugins/directory.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const objectRestSpread = {
+  babelSyntaxPlugin: "syntax-object-rest-spread",
+  babelTransformPlugins: ["transform-object-rest-spread"]
+};
+
+export const asyncGenerators = {
+  babelSyntaxPlugin: "syntax-async-generators",
+  babelTransformPlugins: ["transform-regenerator"]
```
hzoo: we should recommend the new transform, `transform-async-generators`

src/plugins/directory.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const classConstructorCall = {
+  babelSyntaxPlugin: "syntax-class-constructor-call",
+  babelTransformPlugins: ["transform-class-constructor-call"]
```
hzoo: This is deprecated so can remove

src/plugins/directory.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const decorators = {
+  babelSyntaxPlugin: "syntax-decorators",
+  babelTransformPlugins: ["transform-decorators"]
```
hzoo: i guess recommend transform-legacy atm?

src/parser/util.js
```diff
@@ -86,3 +87,25 @@ pp.unexpected = function (pos, messageOrType = "Unexpected token") {
...
+        pointers.push(...pluginInfo.babelTransformPlugins.map(
+          (shortName) => `babel-plugin-${shortName}`
+        ));
+      }
+    }
+    const leadingPointersOr = pointers.length ? ` ${pointers.join(", ")} or ` : " ";
+
+    this.unexpected(null, `This experimental syntax requires the ${name} parser plugin. `
+      + `Check out${ leadingPointersOr }the parserOpts.${name} Babel option.`);
```
hzoo: We should change it so that Babel users will use the transform plugin, and babylon users the syntax, and for Babel users not to use both.

motiz88:
@hzoo so basically - an enhanced error object, whose shape will be part of
Babylon's public API, and which we will then detect and translate on the
Babel side. Right?

My choice for this would either be an ad-hoc extra property e.g.
`missingBabylonPlugin` (and duck typing on that in Babel) or just full
blown VError with its first-class support for metadata. What do you think?

I'll only have my computer back from repairs next week but let's maybe hash
out the details here until I can write actual code.

motiz88:
@hzoo I saw your comment on #300 mentioning this feature. Totally willing to pick this up again. Can you please confirm that the approach I outlined [above](https://github.com/babel/babylon/pull/178#issuecomment-267531497) is how you'd prefer this to be handled? Or do you have something else in mind?

src/parser/util.js
hzoo: Yeah we do have syntax plugins for every transform (otherwise we would just remove the syntax plugins entirely)

hzoo:
Hmm yeah it would be interesting to just expose metadata that says the babylon plugin that could be applied to not error.

motiz88:
Btw just resolved the obvious conflicts for now. Will do a more complete update soon.

src/plugins/directory.js
```diff
@@ -0,0 +1,59 @@
+export const classProperties = {
```
hzoo: Thinking about this again...

most if not all of the plugins have a corresponding babel plugin that is just the same name.

objectRestSpread -> transform-object-rest-spread
asyncGenerators -> transform-async-generators

the only 2 that don't have a corresponding preset now

flow -> preset-flow
react -> preset-react

motiz88:
@hzoo I read "focus on the Babel user" to mean a relaxing of your suggestion above that
> We should change it so that Babel users will use the transform plugin, and babylon users the syntax, and for Babel users not to use both.

So that - showing only the Babel oriented message - would be one simple way to slice it, albeit one that sacrifices some "purity" I guess by coding knowledge about Babel plugins into Babylon (funnily enough, I would be more comfortable with this philosophically if Babylon was in the Babel repo. Ooh can we get that in 7.0? :half_joking:)

src/plugins/directory.js
motiz88: Most is right. There's also the question of `transform` vs `syntax` (even if we ignore the rarer `preset` and `...-legacy` etc) which is still an editorial decision basically

hzoo:
> (funnily enough, I would be more comfortable with this philosophically if Babylon was in the Babel monorepo... Ooh can we get that in 7.0? :half_joking:)

Yeah I've wanted that for a while now (doesn't have to be in 7.0 either but based on time probably so unless someone really wants to do it).

Although I guess it would be a weird error message if the user isn't using Babel but Babylon itself

src/plugins/directory.js
hzoo: -legacy will be removed (there's a pr to make it the regular one) https://github.com/babel/babel/issues/5283

motiz88:
>  it would be a weird error message if the user isn't using Babel but Babylon itself

Yep, that's what I'm least happy about here. I totally get the "focus on Babel users" argument, but Babylon _is_ its own package too and I don't love the idea of making that kind of a second-class story. It's a subtle point though. 

To expand on the alternative: It's more work, but if we make the "plugin required" exception positively identifiable to Babel <sup title="Or anyone invoking Babylon - i.e this becomes public API.">[1](#motiz88-2257-footnote1)</sup>, e.g. via a custom property on the thrown object <sup title="Or subclassing `Error` but fragile; or bringing in `VError` but probably overkill.">[2](#motiz88-2257-footnote2)</sup>, then we can both

1. put the Babylon option name in the Babylon-thrown message, but then catch it in Babel <sup title="In basically just one place, right? `babel-core` somewhere.">[3](#motiz88-2257-footnote3)</sup> and rethrow with the appropriate Babel plugin name;
_and_
2. remove knowledge from Babylon about which Babel plugins exist/are recommended for each syntax feature. ( & move into `babel-core`)

---

<a name="motiz88-2257-footnote1">1</a>: Or anyone invoking Babylon - i.e this becomes public API.
<a name="motiz88-2257-footnote2">2</a>: Or subclassing `Error` but fragile; or bringing in `VError` but probably overkill.
<a name="motiz88-2257-footnote3">3</a>: In basically just one place, right? `babel-core` somewhere.

src/plugins/directory.js
motiz88: Oh yeah, sorry, I was commenting with just a fuzzy recollection of how my own PR worked. So `transform` vs `syntax` is accounted for, by way of both being mentioned in the message.

And now I see your edit above. So yeah, that's a way to encode the same information less explicitly. However I would still kinda prefer to move this out of Babylon if possible, as discussed below.

src/parser/util.js
```diff
@@ -86,3 +87,25 @@ pp.unexpected = function (pos, messageOrType = "Unexpected token") {
...
+  }
+  return true;
```
loganfsmyth: Is there a reason to return from this given that the alternative is to throw? Then we could reduce the nesting by just doing `if (this.hasPlugin(name)) return;` at the top of the function.

src/parser/util.js
```diff
@@ -86,3 +87,25 @@ pp.unexpected = function (pos, messageOrType = "Unexpected token") {
...
+      if (pluginInfo.babelSyntaxPlugin) {
+        pointers.push(`babel-plugin-${pluginInfo.babelSyntaxPlugin}`);
+      }
+      if (pluginInfo.babelTransformPlugins) {
+        pointers.push(...pluginInfo.babelTransformPlugins.map(
+          (shortName) => `babel-plugin-${shortName}`
+        ));
+      }
+    }
+    const leadingPointersOr = pointers.length ? ` ${pointers.join(", ")} or ` : " ";
```
loganfsmyth: How would people feel about skipping this `pointers` logic and having `babylon` just throw an error with the plugin name in the message, and a `.missingPlugin` property?

They we can have the try/catch in `babel-core` catch the exception and handle mapping the error to a Babel plugin? That way we don't need Babylon to be aware of Babel stuff.

hzoo: Yeah that was my thoughts dono if it was in the notes. Babylon would just throw and babel-core would pick that up and throw the real error message since you might be using babylon for something else (prettier, babel-eslint) that is unrelated to using a plugin

PR #182, opened 2016-10-15
Count Babel tests towards Babylon code coverage

motiz88:
As we're already running Babel's tests in CI to detect any integration issues early, we can also use them to meaningfully bump up Babylon's test coverage, by doing this with an instrumented build of Babylon and using codecov's automatic report merging feature.

In practice this currently helps us cover things like `do` expressions and the `::` operator which are conspicuously uncovered in the current `master`. Obviously for new features the ideal still is to add covering tests from the get-go.

(Also, file this under "workarounds we can retire if Babylon is ever merged back into the Babel monorepo" 😉)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
     - node_js: "node"
       env: BABEL=true
 
-after_success: 'if [ -z "${LINT-}" ] && [ -z "${FLOW-}" ] && [ -z "${BABEL-}" ]; then npm run coverage ; fi'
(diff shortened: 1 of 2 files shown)


danez:
Do you think it is okay now that the babel tests are only running on the latest node version? I'm not sure if it should stay like it is or run on all supported node version now that we count it towards the coverage.

motiz88:
@danez I think it's fine as is, assuming these assumptions hold (and I think they do):

1. We only care about hit/miss/partial stats - # of hits per line is irrelevant;
2. We run all the Babel tests at least once;
3. Babel's parsing needs are independent of the version of Node it runs on.

PR #187, opened 2016-10-16
Allow "async" as identifier for object literal property shorthand

### LINKED ISSUE #186: Parsing error under certain scenarios when using `async` as shorthand property in literal
Originally reported by @blesh in https://github.com/babel/babel/issues/4384 - opening an issue here because it looks like a `Babylon` bug.

### Test Case
```javascript
babylon.parse('var foo = {async, bar}') // SyntaxError: Unexpected token (1:16)
```

As pointed out by Ben, this failure doesn't happen in some other situations

### Working examples
```javascript
babylon.parse('var foo = {async}')
babylon.parse('var foo = {bar, async}')
babylon.parse('var foo = {async: async, bar}')
```

DrewML:
Closes #186

### DIFF
--- a/src/parser/expression.js
+++ b/src/parser/expression.js
@@ -765,7 +765,7 @@ pp.parseObj = function (isPattern, refShorthandDefaultPos) {
       if (isGenerator) this.unexpected();
 
       let asyncId = this.parseIdentifier();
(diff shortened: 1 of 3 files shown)


danharper:
:+1: but the test case should include a trailing comma, right?

DrewML:
> 👍 but the test case should include a trailing comma, right?

LOL. I love that I had a test case in the issue and I still made this mistake. Good catch 👍.

I've pushed an update

PR #188, opened 2016-10-16
Error on missing expected.json fixture in CI

motiz88:
This blocks (in a CI environment) the default behavior of automatically creating missing `expected.json` files, thereby guarding against silent failure (and a potential false positive) if a test is accidentally committed without its `expected.json`.


hzoo:
So good!!

PR #190, opened 2016-10-16
Use rollup for bundling to speed up startup time

### LINKED ISSUE #181: Run a bundler (rollup/webpack)
We don't have any dependencies so it sounds like a good idea to do now?

DrewML:
This closes #181. 

Copy/pasting the example I posted in that issue:

**Test case**
```javascript
const babylon = require('babylon');
const ast = babylon.parse('var foo = "lol";');
```

**Without bundling**
image

**With bundling**
image

With that test case, there was a ~95ms savings by removing the need for node to build/traverse the dependency graph.

As noted in that thread, this requires (for now) dropping `npm run watch` due to bugs in `rollup-watch`. Would like to hear feedback from other contributors to determine if this is going to severely impact anyone's workflow.


package.json
```diff
@@ -29,13 +29,16 @@
     "flow-bin": "^0.33.0",
     "lodash": "^4.15.0",
     "nyc": "^8.1.0",
+    "rollup": "^0.36.3",
+    "rollup-plugin-babel": "^2.6.1",
+    "rollup-plugin-node-resolve": "^2.0.0",
```
DrewML: This is required to be able to import an `index` file without having to explicitly list `./foo/index`.

DrewML:
Hmm. Apparently `rollup` only supports node 0.12.x and up. So our output bundle would work on 0.10, but we can't execute the build in Travis for 0.10 (without using something like `babel-register` for that version).

Thoughts?

hzoo:
We could just not run rollup in node 0.10 and do what we were doing before? We should be able to run the tests with/without rollup right?

DrewML:
> We should be able to run the tests with/without rollup right? (might be painful)

My biggest concern there is that, Rollup _could_ theoretically generate a bundle that wouldn't want on 0.10.x, but we wouldn't know if we didn't test the bundle itself.

I did have a thought, though. `nvm` is available on Travis containers. I could have it setup so, for 0.10/0.12, we run the rollup build in node 4, then `nvm use` before the tests run. I'll give it a shot later tonight

wtgtybhertgeghgtwtg:
I doubt this is a feasible suggestion, but since node is dropping support for 0.10 at the end of the month, maybe you could bump Babylon to 7 and make it for node 0.12 and up?  Alternatively, for node 4 and up, since node drops support for 0.12 at the end of the year.

hzoo:
We can just use node 6 to build with rollup and then test on the rest?

danez:
agree,

DrewML:
Cool - I'll push updates to the PR this weekend to build with node latest, then run tests on the target node version

DrewML:
@hzoo @danez Switched things up in travis so builds run in latest node, and then tests are executed in the proper node version for that test run.

Can verify this in the Travis build logs:

image

package.json
```diff
@@ -29,22 +29,24 @@
...
+    "build": "rollup -c",
     "coverage": "nyc report --reporter=json && codecov -f coverage/coverage-final.json",
     "lint": "eslint src bin",
     "flow": "flow",
     "prepublish": "cross-env BABEL_ENV=production npm run build",
     "preversion": "npm run test && npm run changelog",
     "test": "npm run lint && npm run flow && npm run build && npm run test-only",
     "test-only": "ava test",
-    "test-ci": "cross-env BABEL_ENV=test npm run build && nyc npm run test-only",
-    "watch": "babel src --out-dir lib --watch",
```
hzoo: do we have to remove watch if it doesn't need to use rollup?

DrewML: Hmm, you bring up a good point. I just noticed that I didn't change the `test-only` command to run a build first, so technically the code you test against locally isn't 1:1 with the code tested in CI. Thoughts?

hzoo:
```
The 'classCallCheck' Babel helper is used more than once in your code. It's strongly recommended that you use the "external-helpers" plugin or the "es2015-rollup" preset. See https://github.com/rollup/rollup-plugin-babel#configuring-babel for more information
```

Since we already use `modules: false` I guess we can add `  "plugins": [
    "external-helpers"
  ]`

.travis.yml
```diff
@@ -7,7 +7,16 @@ node_js:
...
 
-before_script: 'if [ -n "${BABEL-}" ]; then make bootstrap-babel ; fi'
+before_install:
+    # Rollup doesn't support node < 4.x. Switch to latest for build
+  - . $HOME/.nvm/nvm.sh
+  - nvm install stable && nvm use stable
+
+before_script:
+  - 'if [ -n "${BABEL-}" ]; then make bootstrap-babel ; fi'
+  - 'npm run build'
```
danez: We need to set BABEL_ENV here to test so that babel-plugin istanbul is used.

package.json
```diff
@@ -29,22 +29,24 @@
...
+    "test-ci": "cross-env BABEL_ENV=test nyc npm run test-only",
```
danez: We can remove babel-env here then as it is only used for babel but babel is not run here.

.travis.yml
```diff
@@ -7,7 +7,16 @@ node_js:
   - "5"
   - "6"
 
-before_script: 'if [ -n "${BABEL-}" ]; then make bootstrap-babel ; fi'
+before_install:
+    # Rollup doesn't support node < 4.x. Switch to latest for build
+  - . $HOME/.nvm/nvm.sh
+  - nvm install stable && nvm use stable
```
danez: maybe should we use `nvm install --lts` instead of `nvm install stable` stable is deprecated first of all and is now an alias for node and second it currently installs node 7 and not sure how stable that is already. Or do you think thats okay?

And btw install also automatically calls use.

package.json
danez: I never used watch in babylon as the tests usually do a build before and the build does not take too long.

hzoo:
We can have a separate issue to use `external-helpers` - maybe it's a babel issue or a rollup one not sure but it's including the newly added asyncGenerator helper for some reason which makes it bigger https://gist.github.com/hzoo/6f393d0a24be191e25dee0bb894bb09e

Wonder how hard it would be to just dedupe it in a plugin instead

kittens:
Yay, this will also stop people doing dumb shit like [this](https://medium.com/@jacobp100/adding-custom-syntax-to-babel-e1a1315c6a90#.4rzzuzqtb).

Issue #134, opened 2016-09-20
Fix reserved keywords


danez:
http://www.ecma-international.org/ecma-262/6.0/#sec-keywords

Some throw correctly: `for`, `implements`, ...

- [x] Some don't: `enum`, `await`
- [ ] Some are supposed to be usable in script mode (non-strict) but don't: `let` (need to test other envs if this is true)

This list is probably not complete.

kaicataldo:
I'm interested in looking at this! I have a few other commitments but can try to look at this soon. If someone else has the bandwidth right now though, please feel free to take it on.

danez:
Awesome.
This issue contains 2 issues I think (throw for invalid identifier and the let keyword which should be usable in script mode, although not sure for the last). Feel free to only fix one of them first if its easier.

hzlmn:
@kaicataldo is it still an issue?

kaicataldo:
Yes! I started looking at the `let` in strict mode issue, but haven't started on the others.

hzlmn:
@kaicataldo Can I look on `await` or `enum` ?

kaicataldo:
@hzlmn Absolutely - don't want to be a blocker. Thanks for looking at this :)

hzlmn:
@kaicataldo start looking here https://github.com/hzlmn/babylon/commit/2498481a511aaaf50d3f5ab61528e2d81910d486
. Can you look on this, so I will not bother with silly PRs?

kaicataldo:
@hzlmn I recommend starting with some tests, since you know the desired result. It'll help guide what you do a lot!

kaicataldo:
I just realized another issue I'm working on has some overlap with this one, so I should be covering the `await` portion of this task

hzlmn:
@kaicataldo No problems, if you already done it. I am pretty new to `babylon` codebase, so even simple task like this, can take a while.

kaicataldo:
As a quick update on the state of this issue: 

The `enum` and `await` part has been taken care of. However, the second part regarding `let` in script mode still needs to be done.

mannanali413:
@kaicataldo is the issue with the ``` let ``` keyword resolved. If not then I can start working on it.

kaicataldo:
@mannanali413 Please feel free!

PR #194, opened 2016-10-22
WIP: Better error messaging for incorrect use of 'await'

kaicataldo:
Fixes https://github.com/babel/babylon/issues/113

This PR does two things:
 * Makes it so that the user sees a helpful error message when `await` is not used in an `async` function
 * Disallows the use of `async` as an `Identifier` when source type is module (as outlined in https://github.com/babel/babylon/issues/134). See http://www.ecma-international.org/ecma-262/6.0/#sec-future-reserved-words for relevant part of the spec.

Marking this as WIP because I haven't been able to figure out how to warn for `await` outside of functions safely. Please see comment below.


test/fixtures/es2017/async-functions/44/options.json
```diff
@@ -0,0 +1,3 @@
+{
+  "throws": "Unexpected token, expected ; (1:6)"
```
kaicataldo: I'm still unsure how we can safely warn insert a better error message for this case, given that this needs to be allowed: https://github.com/babel/babylon/blob/master/test/fixtures/esprima/es2015-identifier/valid_await/actual.js

kaicataldo:
Was just made aware of https://github.com/babel/babylon/pull/125. We can just close this if that seems like the better way to go :)

PR #195, opened 2016-10-24
Throw error for reserved words enum and await

kaicataldo:
Refs https://github.com/babel/babylon/issues/134

**Removed check and tests for renamed destructuring assignments. See comments below.**
~~This PR ended up being a bit bigger because Babylon currently doesn't check renamed destructured values to see if the new name is a reserved word/keyword. So before this PR, the following would not warn:~~

``` js
function foo({ bar: enum }) {}
function foo({ bar: protected }) {} // in strict mode
const { foo: enum } = bar();
const { foo: protected } = bar(); // in strict mode
```

~~To ensure the reserved words error is thrown everywhere, I also implemented this check. Is this too many changes in one PR?~~


kaicataldo:
Looks like I might also need to write some more tests to get higher coverage

hzoo:
I think we should make the `check renamed destructured values` part as a separate PR since there's a lot of changes to review and they can be independent. Easir to merge both then?

And maybe consolidate the error messages so it's the basically the same thing - not being able to use reserved words and in strict mode or not

kaicataldo:
@hzoo How do these changes look? I removed the check for renamed destructuring assignments and centralized the reserved word checking logic. Wanted to double check before fixing all the tests (129 failing tests now because the error message has changed to the more generic one) :)

I definitely think this will be easier to maintain :+1:

src/parser/expression.js
```diff
@@ -827,13 +827,11 @@ pp.parseObjPropValue = function (prop, startPos, startLoc, isGenerator, isAsync,
 
   if (!prop.computed && prop.key.type === "Identifier") {
     if (isPattern) {
-      let illegalBinding = this.isKeyword(prop.key.name);
-      if (!illegalBinding && this.state.strict) {
-        illegalBinding = reservedWords.strictBind(prop.key.name) || reservedWords.strict(prop.key.name);
-      }
-      if (illegalBinding) {
-        this.raise(prop.key.start, "Binding " + prop.key.name);
+      if (this.isKeyword(prop.key.name)) {
```
kaicataldo: I guess the case slips through the tokenizer check for `isKeyword`? Leaving this because there's one extra failing test when it's removed.

src/parser/expression.js
```diff
@@ -827,11 +827,7 @@ pp.parseObjPropValue = function (prop, startPos, startLoc, isGenerator, isAsync,
 
   if (!prop.computed && prop.key.type === "Identifier") {
     if (isPattern) {
-      if (this.isKeyword(prop.key.name)) {
```
kaicataldo: Moved this into the `checkReservedWord()` method, since it should be with the other checks.

kaicataldo:
Updated! Let me know if there's anything else that needs to be done. Will follow this PR up with another PR that uses the `checkReservedWords()` method to check for reserved words for renamed destructed assignments (since those are not currently checked).

kaicataldo:
@danez Thoughts on this approach?

PR #197, opened 2016-10-25
Optional names for function types and object type indexers

gabelevi:
In Flow v0.34.0, Flow will no longer require function type params and object type indexers to have names. So where you used to have to write

```
type A = (x: string, y: boolean) => number;
type B = (z: string) => number;
type C = { [key: string]: number };
```

you can now write

```
type A = (string, boolean) => number;
type B = string => number;
type C = { [string]: number };
```

Note: If you want to use a function type as the return type for an arrow function expression, you must either wrap it in parens or name each param:

```
let f1 = () : (string => number) => ...; // OK due to parens
let f2 = (): ((string) => number) => ...; // OK due to parens
let f3 = (): (x: string) => number => ...; // OK due to named param
let f4 = (): string => number => ...; // Error!
```


.gitattributes
```diff
@@ -1 +1,3 @@
 * text eol=lf
+test/fixtures/es2015/uncategorised/22/actual.js binary
+test/fixtures/es2015/uncategorised/23/actual.js binary
```
danez: What was the issue you had with these two files?

src/plugins/flow.js
```diff
@@ -529,6 +552,7 @@ pp.flowParsePrimaryType = function () {
   let tmp;
   let type;
   let isGroupedType = false;
+  let oldNoAnonFunctionType = this.state.noAnonFunctionType;
```
danez: Can you please add `noAnonFunctionType` to the [state](https://github.com/babel/babylon/blob/master/src/tokenizer/state.js) so that v8 can better optimize the usage of the state.

test/fixtures/flow/anonymous-function-no-parens-types/bad_01/actual.js
```diff
@@ -0,0 +1 @@
+var f = (x): number => 123 => 123;
```
danez: It would be nice to have a better error message that says something like `return type annotation needs to be wrapped in parens or have named parameters`. But not sure if this is easy to accomplish.
This should not be a blocker, just if you have an idea on how this could be done easy, otherwise we can come back to that in a separate story.

danez:
Thanks for the PR, have just on micro performance change that would be nice to do, and 2 questions.
👍

.gitattributes
gabelevi: I think they have weird characters that the `.gitattribute` doesn't like if it treats them as text

```
$ git checkout master .gitattributes
$ git commit -m "revert gitattributes"
$ git rm --cached -r .
$ git reset --hard
$ g st
On branch flow
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   test/fixtures/esprima/invalid-syntax/migrated_0155/actual.js
	modified:   test/fixtures/esprima/invalid-syntax/migrated_0159/actual.js

no changes added to commit (use "git add" and/or "git commit -a")
$ g diff
warning: CRLF will be replaced by LF in test/fixtures/esprima/invalid-syntax/migrated_0155/actual.js.
The file will have its original line endings in your working directory.
warning: CRLF will be replaced by LF in test/fixtures/esprima/invalid-syntax/migrated_0159/actual.js.
The file will have its original line endings in your working directory.
diff --git a/test/fixtures/esprima/invalid-syntax/migrated_0155/actual.js b/test/fixtures/esprima/invalid-syntax/migrated_0155/actual.js
index dd8e38b..ed49ef3 100644
--- a/test/fixtures/esprima/invalid-syntax/migrated_0155/actual.js
+++ b/test/fixtures/esprima/invalid-syntax/migrated_0155/actual.js
@@ -1,2 +1,2 @@
-//
+//
 ]
diff --git a/test/fixtures/esprima/invalid-syntax/migrated_0159/actual.js b/test/fixtures/esprima/invalid-syntax/migrated_0159/actual.js
index 583aa13..e85031c 100644
--- a/test/fixtures/esprima/invalid-syntax/migrated_0159/actual.js
+++ b/test/fixtures/esprima/invalid-syntax/migrated_0159/actual.js
@@ -1,2 +1,2 @@
-/*
+/*
 */]
```

src/plugins/flow.js
gabelevi: Sure, will do!

test/fixtures/flow/anonymous-function-no-parens-types/bad_01/actual.js
gabelevi: Yeah, it's not easy to do. The problem is that the first `=>` is parsed as arrow function's `=>`. So we'd have to decide which parsing errors after the `=>` would need this explanation and which are just normal arrow function parse errors.

gabelevi:
Can this be merged now?

danez:
Sorry I was on vacation and just got back, can merge tomorrow and try to get a release soon.

Issue #217, opened 2016-11-11
Flow type syntax error for nested arrays


bxt:
I opened this in response to the upstream bug babel/babel#4831

### Input Code
Using babylon `6.13.1` this code with a type annotation for a nested array produces an error:

```js
const foo: number[][] = [[]];
```

```js
require("babylon").parse("const foo: number[][] = [[]];", {
  // parse in strict mode and allow module declarations
  sourceType: "module",

  plugins: [
    "flow"
  ]
}
```

```
/node_modules/babylon/lib/index.js:4218
  throw err;
  ^

SyntaxError: Unexpected token (1:19)
    at Parser.pp$5.raise (/node_modules/babylon/lib/index.js:4215:13)
    at Parser.pp.unexpected (/node_modules/babylon/lib/index.js:1599:8)
    at Parser.pp$1.parseVar (/node_modules/babylon/lib/index.js:2179:12)
    at Parser.pp$1.parseVarStatement (/node_modules/babylon/lib/index.js:2008:8)
    at Parser.pp$1.parseStatement (/node_modules/babylon/lib/index.js:1701:19)
    at Parser.parseStatement (/node_modules/babylon/lib/index.js:5130:22)
    at Parser.pp$1.parseBlockBody (/node_modules/babylon/lib/index.js:2105:21)
    at Parser.pp$1.parseTopLevel (/node_modules/babylon/lib/index.js:1617:8)
    at Parser.parse (/node_modules/babylon/lib/index.js:1509:17)
    at Object.parse$1 [as parse] (/node_modules/babylon/lib/index.js:6371:37)

```

Parsing this equivalent code works perfectly fine:

```js
const foo: Array<number[]> = [[]];
```

Flow parses both of them just fine, as you [can see here](https://flowtype.org/try/#0MYewdgzgLgBAZiEAuGYCuBbARgUwE4DaAusTALwwHFEDcAUHaJLFgIZ4oCCeerAngB502fMQB85StXpA).

### Babel Configuration

`.babelrc` (e.g.):

```js
{ "presets": ["es2015"], "plugins": ["transform-flow-strip-types"] }
```

`package.json` (e.g.):

```js
{
  "main": "index.js",
  "dependencies": {
    "babel-cli": "^6.18.0"
  },
  "devDependencies": {
    "babel-plugin-transform-flow-strip-types": "^6.18.0",
    "babel-preset-es2015": "^6.18.0"
  },
  "scripts": {
[...]

PR #219, opened 2016-11-12
Parse flow nested array type annotations like number[][]

bxt:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes?
| Tests added/pass? | yes
| Fixed tickets     | #217 
| License           | MIT

Flow supports nested array types like `number[][]`, while babylon throws a syntax error, as noted in #217 . This error caused problems upstream, e.g. when trying to strip the types with babel, see babel/babel#4831. The workaround was to use `Array<number[]>`. To solve this, I adjusted the `flowParsePostfixType` method to parse 0-n `[]` postfixes and return the correct AST. I added test cases with:

```
var a: number[][]
```
```
var a: number[][][]
```

For the tests' `expected.json`s I copied flow's ASTs from astexplorer.net and adjusted the `range` keys etc. to match babylon's format. I was curious if there could be any interference with JS arrays, so I checked with [the code of flow's parser](https://github.com/facebook/flow/blob/ef82d3d8cdf78ca2b7e532e55498454e89a75b45/src/parser/parser_flow.ml#L204). And indeed – they check for line terminators there, so I added another test case with:

```
var a: number
[]
```

And made sure this parses as a program with a body of a `VariableDeclaration ` and an `ExpressionStatement`.

Let me know what you think. ☺️


danez:
Awesome, thanks for the contribution. Looks good, but I'm currently on my phone will check later when I'm home.

hzoo:
Awesome contribution @bxt (also your issue description was beautiful)

bxt:
@danez @hzoo Thank you ☺️ nice to work with you.

PR #225, opened 2016-11-13
Run tests of flow with babylon

danez:
This was a try to run all tests from the flow repository with babylon.
Currently 100 of 590 fail, although there could be still false positives.

What are your opinions on that? We can probably never safely run them without allow_failures, but I thought it might be a nice hint on what are the differences between parsing in flow and babylon.


hzoo:
Sounds cool to me - more tests the better?

PR #227, opened 2016-11-16
use preset-env in tests

hzoo:
Wanted to test out using preset-env on a project..

leaving `"presets": ["stage-0"]` at the top level caused an issue with "missing class properties transform" since I believe the es2015 preset ran before class properties due to env and merging which is unfortunate.

Also means that we're duplicating the config for the different envs

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,17 +1,29 @@
 {
-  "presets": [
-    ["es2015", {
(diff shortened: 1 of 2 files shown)


hzoo:
Nvm forgot rollup was node >= 4

PR #228, opened 2016-11-17
Change location of ObjectTypeIndexer to match flow

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | no
| License           | MIT

```js
interface Dictionary { [index: string]: string; length: number; }
```

http://astexplorer.net/#/TtDjoZgl57/1

The location of the `ObjectTypeIndexer` is currently different in flow and babylon. Babylon includes the semicolon, whereas flow does not. 

Usually in all parsers for ES-Code the semicolon/colon is always included, like in ClassProperties and ObjectProperties. 

So I'm not sure if we should do this, because it is not super consistent, but it would be inline with the flow parser then.

//cc @samwgoldman @gabelevi @mroch


samwgoldman:
We can also change the Flow parser to be consistent with Babylon. WDYT @mroch?

danez:
I'm going to merge this for 7.0. If you still want to change flow at some point I'm happy to revert it, but I think it is more important right now to be more consistent with the flow parser.

PR #229, opened 2016-11-17
Allow plugins:["*"]

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | no
| Fixed tickets     | no


hzoo:
A test using `*` would be good 😄

hzoo:
a bug fix since it should of worked before?

krainboltgreene:
Does this mean I can automatically load all plugins I have installed? How does it figure out which package is a plugin?

DrewML:
@krainboltgreene Bablyon doesn't support external plugins (so _plugin_ is a bit of a misnomer here). The `plugins` option takes an array of strings for _built-in_ parser plugins.

hzoo:
Right this is only for the parser plugins https://github.com/babel/babylon#plugins - it's basically a truthy check in the parser to either make it say "unexpected token" or to parse it.

PR #234, opened 2016-11-24
Add watch script for dev

kaicataldo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | no
| License           | MIT

The watch script was removed in [this PR](https://github.com/babel/babylon/commit/beb8db62646081a88b8db31fe95d4b37aa28d98f) when we started bundling Babylon with rollup (which is awesome!). This PR sets up a watch script for us (the big difference is that with Rollup the `es2015` preset's `modules` option must be `false`, while we need to enable transpiling to CommonJS modules when Babel is run).

 I tried Rollup's watcher, but [it is currently WIP](https://github.com/rollup/rollup-watch), and would crash and was unable to recover from saving files with syntax errors when I tried it.

This is more of a stop gap measure until the Rollup watcher is ready for primetime.

Thoughts on this? I personally was using the watcher and have continued to set up something like this up locally when I'm working on Babylon, so figured it might be nice to have a watch script again. Totally understand we probably want to use one of the two solutions above, but I figured having a watcher is better than not :)


danez:
It would be nice in general that the tests maybe not use the rollup bundle, so that we can have proper unit tests, right now the only thing that can be tested is the public API.
And then use rollup only for publishing. Although it is probably also nice to know that the rollup bundle works as expected.

danez:
I changed it so that the tests run without rollup.
@DrewML Do you think this is okay?

.babelrc
```diff
@@ -1,16 +1,32 @@
...
+    "production": {
+      "presets": [
+        ["es2015", {
+          "loose": true,
+          "modules": false
+        }],
+        "stage-0"
+      ]
+    },
+    "development": {
```
kaicataldo: I wonder if we should just combine `development` and `test, since they're the same? Maybe just `production` and `development?

kaicataldo: Oh I see, it's for the `plugins`. Never mind!

DrewML:
> I changed it so that the tests run without rollup.
@DrewML Do you think this is okay?

I think the only part that would be concerning is not running any tests against the bundle in CI. It's possible `rollup` could introduce a bug that we would never catch until running the `Babel` test suite against a release of `Babylon`.

kaicataldo:
> Thinking about it more, we run the Babel test suite in CI against each Babylon PR, right?

Yeah! Seems like we could run that CI test against the bundle?

> It would be nice in general that the tests maybe not use the rollup bundle, so that we can have proper unit tests, right now the only thing that can be tested is the public API.

Could it make sense to split this out into unit and integration/behavioral tests and not bundle for the first and then bundle for the latter?

kaicataldo:
Maybe it would make more sense to separate these two tasks out? Seems like we could just do the watcher script in this one and we then have the env's to take it further for testing, etc. in a future PR?

hzoo:
I think @kaicataldo just wanted to be able to watch in dev as this PR. If it's simple enough we can just land that and then work on the other stuff. Also keep in mind we probably want to just move it back into the monorepo as well 😝

danez:
Yeah true, feel free to reset the branch to your commit and force push.

kaicataldo:
Updated to just add a watch script

.babelrc
```diff
@@ -10,6 +10,14 @@
     "transform-flow-strip-types"
   ],
   "env": {
+    "watch": {
+      "presets": [
+        ["es2015", {
```
hzoo: might be able to just do plugins commonjs here due to the inheriting of the top level presets/plugins but this works fine

hzoo:
👍

.babelrc
kaicataldo: Ohh yeah that could be better :+1:

Issue #231, opened 2016-11-21
Bug - decorator with computed name at end of class


arackaf:
re-opening this case 

https://github.com/babel/babel/issues/4866#event-865830403

here by request.

It looks like Babel is erroring out when there's a decorated class property method as the last statement in a class.

```js
const mixin = options => Class => class sub extends Class {
    @action [options.foo] = () => true;
};
```

Which produces

> SyntaxError ......  You have trailing decorators with no method (68:1)

Adding a junk placeholder at the end fixes it, as a workaround

```js
const mixin = options => Class => class sub extends Class {
    @action [options.foo] = () => true;

    __junk;
};
```

### Babel Configuration (.babelrc, package.json, cli command)

```js
{
  "name": "transpiler",
  "version": "1.0.0",
  "description": "",
  "author": "",
  "license": "ISC",
  "dependencies": {
    "babel-plugin-transform-decorators-legacy": "1.3.4",
    "babel-preset-es2015": "6.18.0",
    "babel-preset-react": "6.16.0",
    "babel-preset-stage-2": "6.18.0",
    "del": "2.2.2",
    "gulp": "3.9.1",
    "gulp-babel": "6.1.2",
    "gulp-if": "2.0.0",
    "gulp-line-ending-corrector": "1.0.1",
    "gulp-notify": "2.2.0",
    "gulp-plumber": "0.6.6",
    "gulp-print": "1.1.0",
    "gulp-rename": "1.2.0",
    "jest": "15.1.1",
    "jest-cli": "15.1.1"
  },
  "jest": {
    "testPathDirs": ["test"],
    "modulePaths": ["2"],
    "testRegex": ".js$",
    "moduleNameMapper": {
      "^react$": "assets/react/react",
      "^react-dom$": "assets/react-dom/react-dom",
      "^mobx$": "assets/mobx/mobx",
      "^mobx-react$": "assets/mobx-react/index.js",
      "^bootstrap-dropdown-button$": "assets/react-bootstrap/buttonDropdown.js"
    },
    "testPathIgnorePatterns": [
      "test/mocks/"
    ]
  },
  "scripts": {
    "preinstall": "npm install -g gulp@3.9.1",
    "test": "jest"
  }
}
```

### Your Environment

| software         | version
| ---------------- | -------
| Babel            |  see above
| node             | 4.4.5
| npm              | 3.9.5
| Operating System | Windows 10

DrewML:
[...]

PR #236, opened 2016-11-25
WIP: Add new plugin (classAndFieldDecorators) for Stage-2 decorators proposal

DrewML:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #232, #231
| License           | MIT

This is one of several PRs that will be coming in (eventually) to start updating parsing to the latest draft of the decorators proposal (Stage 2).

[Proposal is here](http://tc39.github.io/proposal-decorators/).

The biggest change of the new latest proposal introduces a new type of `MemberExpression` production (`DecoratorMemberExpression`), that does not allow computed property access (`[]`). This was added to address the ambiguity between a computed member expr and a computed class property.

```
DecoratorMemberExpression[Yield]:
    IdentifierReference[?Yield]
    DecoratorMemberExpression[?Yield].IdentifierName
```

**TODO**
- [ ] Add `hasPlugin` check for `classAndFieldDecorators` everywhere that a check for the `decorators` plugin is, that is _also_ compliant with the latest revision of the proposal.
- [ ] Move tests for existing decorators functionality to `decorators-legacy` test folder
- [ ] Disallow computed member expressions or decorated computed properties in obj literals


loganfsmyth:
Will this be breaking for cases like

```
class Foo {
    @decorators['foo']
    bar(){}
}
```
?

DrewML:
> Will this be breaking for cases like

It _should_ be, based on my reading of the spec (good catch, updated to breaking in the OP). Having said that, this is a case that this branch currently parses incorrectly. I'll work on fixing that.

loganfsmyth:
Breaking changes are something we need to figure out in Babel and Babylon. This kind of thing makes me wonder if we should add a `decorators2016` plugin and leave the current one as-is or something, since we can't really make breaking changes to the existing logic and bumping to 7.x would have its own set of downsides.

DrewML:
How about `classAndFieldDecorators`? That way we're not tying it to a year.

The other question, though, would be what should we do in the scenario that a user has both enabled (could easily happen now that we support `*` for all plugins)? With your example code above, one version of the draft (the older one) would allow `@decorators['foo']` to parse, but the newest draft would consider that a syntax error.

danez:
Should we make a new major version and remove the old decorator plugin and instead create the new ones? I think we also have other stuff that is breaking. node 4, deprecated plugins

peey:
I think this PR is incorrect because

 1. `@decorators['foo']` (is allowed)[https://github.com/babel/babylon/pull/236/files#diff-1dfdb9c8d46a6ff5fbe5c1e2bb261ed3R1]
 2. [It doesn't allow decorating computed properties](https://github.com/babel/babylon/pull/236/files#diff-71f5ec1ce64b229d1ebbcf5d9a486d8aR673)

If my reading of the spec is correct, then for 1 only `@foo.bar.baz` or `@foo.bar.baz(a,b,c)` syntax should be allowed where `foo,bar, or baz` can be any `IdentifierName` and decorating computed properties should be legal

hzoo:
Going to close in favor of https://github.com/babel/babylon/pull/587!

PR #240, opened 2016-11-28
Prepare tests for multiple fixture runners.

danez:
This refactor the test file, and extracts the actual code into a util file, so that we can create a second fixture runner in future. This is a preparation for #213 

The only codechange is the remove of lodash and the usage of native iterators.


jeromew:
@danez thanks for this PR. I was looking into how I could add tests for #213 and was going to ask if I could add a separate fixture directory for this ;-)
I'll wait until this PR is merged. Can you tell me on #213 what do you think the name should be for the fixture dir and a bit about the coverage that I should try to reach in the PR ?

Issue #238, opened 2016-11-26
`export new Foo()` shouldn't parse


zertosh:
Parses with babylon@16.4.1, but doesn't with acorn@4.0.3, esprima@3.1.1 or flow@0.36.0.

A few more examples:

```js
export typeof foo;
export void 0;
export delete foo.bar;
export this.foo;
```

Seems like an issue with any unary operator.

DrewML:
Another example: statements

```js
export for (;;);
```

PR #241, opened 2016-11-29
Throw error when exporting non-declaration

kaicataldo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #238
| License           | MIT

Currently, Babylon will parse any export that starts with a keyword as a declaration. These changes will throw when the statement following the `export` keyword is not a declaration.


src/parser/statement.js
```diff
@@ -861,8 +861,22 @@ pp.parseExport = function (node) {
...
+  return node.type === "VariableDeclaration"
+    || node.type === "FunctionDeclaration"
+    || node.type === "ClassDeclaration";
+};
+
 pp.parseExportDeclaration = function () {
-  return this.parseStatement(true);
+  const statement = this.parseStatement(true);
+  if (!this.isDeclaration(statement)) {
+    this.unexpected(statement.start);
```
DrewML: Since we know what the problem is here (an `export` declaration that doesn't have a valid export declaration or export clause) maybe we can use `this.raise` and hint to the user what the problem was?

kaicataldo:
Hmmm, actually, going to explore checking token values ahead of time rather than the node type afterwards

src/parser/statement.js
kaicataldo: After discussing a bit with @hzoo, I changed it to an ahead of time check rather than checking the export statement post-parse. This actually changes the error message in some cases, so wanted to see what do you think. 

If we want a more descriptive error message I could leave the original conditional and do the `shouldParseExportDeclaration` check (maybe now called `isDeclaration`?) inside that block, throwing a more descriptive error if `isDeclaration` returns `false`.

My only concern with the error message is that the intention of the code is. If Babylon is trying to parse `export typeof foo`, can we surmise they intended to export a declaration? Or did you mean a more generic `Not a valid export` type of message? Let me know if I'm missing something!

kaicataldo:
Updated

kaicataldo:
Sorry, one last update - added return value flow type and moved the `shouldParseExportDeclaration` method back to where it was before I removed the other methods

danez:
Nice thanks

PR #243, opened 2016-11-29
Use fromCodePoint to convert high value unicode entities

ryanjduffy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| License           | MIT

High code point Unicode entity references are being truncated because `String.fromCharCode` can only return a 2-byte character whereas `String.fromCodePoint` can return a 4-byte character. Interestingly, the existing test case includes an appropriately high code point reference but the expected result is the truncated value and not the truly expected value.

With this change, `&#x1f4a9;` is correctly converted to `💩` instead of `` (which would be the value of `&#xf4a9;`).


danez:
Thanks for spotting this. Awesome.

You added the polyfill as devDependency although it probably should be a dependency. But even then it won't work I think, because we use rollup to bundle babylon.

Other option would be to drop support for node < 4.
@hzoo @DrewML ideas?

hzoo:
Wouldn't rollup just bundle the polyfill in?

Yeah we can drop it but not in this pr, would need to plan that out

danez:
not sure how the bundling works, would need to check.

hzoo:
Oh - then we need to use a version that gives a reference rather than the actual polyfill - like this kind of thing

```js
var fromCodePoint = require('fromCodePoint');
fromCodePoint();
```

ryanjduffy:
If the plan is to drop support for node < 4 anyway, what do you think of including an inline copy of the polyfill (the fn without updating `String`, that is)? The polyfill code hasn't been updated in 2.5 years so there's not much risk of getting out of sync between now and when you'd be able to remove it. Seems like a safe and minimally intrusive alternate.

ryanjduffy:
Went ahead and made a copy of the polyfill and dropped the dependency. Made the necessary changes to adapt it to the babylon lint settings and avoid updating `String`. Let me know what you think.

cc: @mathiasbynens - fyi on copying String.fromcodepoint (and thx for the polyfill!).

mathiasbynens:
LGTM! Thanks for the CC, @ryanjduffy — I appreciate the heads-up :)

Ref. #4315 for dropping Node.js < 4 (in which case the polyfill can be removed).

mathiasbynens:
Btw, since the goal is to decode HTML entities, consider using a library like [he](https://github.com/mathiasbynens/he).

PR #256, opened 2016-12-13
flow: allow leading pipes in all positions

vkurchatkin:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

/cc @jeffmo


danez:
Is that a new change in flow? OR was that wrong in babylon?
Anyway if it should be really allowed everywhere a type can appear, then this looks good to me.

vkurchatkin:
I don't think it's new and have no idea why it wasn't implemented like this in the first place. I hope @jeffmo could clarify.

I've tested all examples with Flow and they work fine.

vkurchatkin:
I've taken a look at Flow's source code and that's exactly what they do

jeffmo:
Sorry for the delay -- it was extended to any type annotation (rather than just aliases) back in v0.31.
A quick look at the tests here seem legit

PR #260, opened 2016-12-18
Modify grammar to support Private Fields proposal:

diervo:
First step to support [Private Fields](/babel/babel/issues/4408)
Implemented grammar changes according to the [spec](https://github.com/tc39/proposal-private-fields):
- Adding a new (optional) plugin `classPrivateProperties`
- Adding [PrivateName](https://tc39.github.io/proposal-private-fields) type identifier
- Adding ClassPrivateProperty to ClassBody
- Allow PrivateName in MemberExpression
- Allow PrivateName as a [references](https://tc39.github.io/proposal-private-fields/#sec-private-references)

_Bonus_: Minor renaming and small code tweaks towards a future refactor within ClassBody parsing method.


src/parser/expression.js
```diff
@@ -1014,6 +1022,15 @@ pp.parseIdentifier = function (liberal) {
   return this.finishNode(node, "Identifier");
 };
 
+pp.parsePrivateName = function (liberal) {
+  const isPrivate = this.eat(tt.hash);
+  const node = this.parseIdentifier(liberal);
```
littledan: Should this be parsePropertyName?

src/parser/statement.js
```diff
@@ -652,62 +652,72 @@ pp.parseClassBody = function (node) {
...
-      method.decorators = decorators;
+      elementNode.decorators = decorators;
       decorators = [];
     }
 
+    if (this.match(tt.hash)) { // Private property
+        if (!this.hasPlugin("classPrivateProperties")) {
+          return this.raise(this.state.pos, `Unexpected character '${tt.hash.label}'`);
+        }
+        this.next();
```
littledan: Will this ignore whitespace? In the current spec, whitespace between the # and the IdentifierName is not allowed (though Waldemar proposed that it be allowed).

ast/spec.md
```diff
@@ -149,6 +151,15 @@ interface Identifier <: Expression, Pattern {
 }
 ```
 
+# PrivateName
+
+```js
+interface PrivateName <: Expression, Pattern {
```
loganfsmyth: I'm curious to hear other's thoughts, but I'd probably have expected this to be a normal `Identifier`, rather than having a specific subtype for the private now.

littledan: The grammar of PrivateNames is pretty different from Identifiers, in terms of where they can be used. It's been suggested that the grammar in the specification be refactored to not call it PrivateName but more like PrivateReference, consisting of the two individual tokens. To me, it makes sense that they are treated differently from identifiers.

src/parser/statement.js
diervo: I will test it and probably I should write a test for it.

ast/spec.md
diervo: Apart from dan's comments about grammar, @loganfsmyth, I though it was cleaner (at least while is at Stage2) to make additive changes when possible rather than mutating existing nodes, like adding a `private` attribute for this case. We can recondile in the next mayor version, once public fields, private fields et al, are in +Stage3

diervo:
Someone needs to help me figure out what to do with the one test failure.

DrewML:
The test failure looks like it is catching something valid-ish.

> ✖ esprima/invalid syntax/migrated_0065 Error: /home/travis/build/babel/babylon/test/fixtures/esprima/invalid-syntax/migrated_0065/actual.js: Expected error message: Unexpected character '#' (1:2). Got error message: Unexpected token, expected ; (1:2)

For that particular test-case, it seems like the error we're spitting out now is much less descriptive of the problem.

ast/spec.md
xtuc: #### Identifier

It is an identifier with a specific type.
It would be confusing for developers because they will be forced to check the type of the identifier.

Most of them will omnit it and will cause some bugs.


#### PrivateName

More explicit.

I don't have much context here, but I guess most transformation within a class will be with both PrivateNames and Identifier.

src/parser/expression.js
littledan: I'm not talking about naming convention, I'm talking about grammar. The grammar for Identifier is more restrictive than IdentifierName/property names, in that it doesn't allow keywords. The grammar for private state specifically allows keywords as the name (just like ordinary properties).

diervo:
I have fixed the test issue due to an less descriptive message in the test that runs `id #= 1`, but in order to do that I added a check on parseMaybeAssign ( see code changes here https://github.com/babel/babylon/pull/260/commits/beef2ef8f4cfaec19b35c44c5435e314889717d8).

Although that does work, it prevents for example to use the decorators since it will throw because parseMaybeAssign is used inside decorators as well. The best solution I found was to add in parseMaybeAssign another flag to detect when to raise the error, but I don't like it.

I'm not very happy with this change.
Thoughts?

diervo:
Alright this PR has been dead for way too long, and I promised @littledan I will try to land it before the next TC39 meeting :)

I fixed the couple of outstanding test issues and added specific more tests for private fields.

There are two small details based on the feedback here and the conversation on [estree #154](https://github.com/estree/estree/issues/154):

First is about internal naming convention for `ClassProperty` and `ClassPrivateProperty`. I just named the new Node based on the previous naming convention for now. Given than the committee was strongly in favor of unifying both proposals going forward, we can normalize the names when both proposals reach Stage3 (I volunteer to do the work).

The second issue is about wether PrivateName contains in its name a `string` or an `Identifier` (more specifically a `namedIdentifier`). Again this are mostly implementation details (although this one matters towards the integration with babel and the transforms). I will defer to @littledan for weight on this one.

IMHO any of the previous points shouldn't be blocking this PR, otherwise we will continue to be stacked without being able to get feedback from developers and push this hopefully forward.

@loganfsmyth @hzoo Can we do another pass at this PR, and unless there is other concerns, merge it?

littledan:
I'm fine with the current code. My only nitpick would be, maybe there doesn't need to be a 'liberal' parameter, as private names are always liberal.

diervo:
@littledan suggestion done!
@hzoo This code coverage thing, doesn't make too much sense to me...
What is remaining to merge this PR?

hzoo:
I think not much if we are ok with the naming conventions - I think we should be ok with making the breaking changes if necessary. You can ignore the coverage issue if it's not making sense 

However (I guess this is my responsibility not yours 😄 ) wanted to be clearer about how we do proposals moving forward (meta issue). I thought that by making a "plugin" like flow/jsx do we could have independent versions of proposals.

So if a breaking change is necessary (and it will probably be if we change the AST node names), how can we update the parser without requiring a major version bump? The proposal was to just make another plugin like "classPrivateProperties2" or something etc and then it would be a minor version bump in babylon itself while the transform could do a major version.

diervo:
@hzoo Ready to be versioned in the future :)

diervo:
@hzoo Who else I need to bribe to get this in? :)

ast/spec.md
```diff
@@ -166,6 +168,18 @@ interface Identifier <: Expression, Pattern {
 
 An identifier. Note that an identifier may be an expression or a destructuring pattern.
 
+
+# PrivateName
+
+```js
+interface PrivateName <: Expression, Pattern {
+  type: "PrivateName";
+  name: string;
```
loganfsmyth: Still feeling like this should be `Identifier`, otherwise we're diverging from the semantics that are used throughout the rest of the AST, where literal text blocks without quotes are `Identifier` nodes.

I agree it's not as pretty nesting-wise, but it seems more consistent.

src/tokenizer/index.js
```diff
@@ -407,6 +407,13 @@ export default class Tokenizer extends LocationParser {
     switch (code) {
       // The interpretation of a dot depends on whether it is followed
```
Qantas94Heavy: I think this comment should go after the new case for 35 (`"#"`).

src/parser/expression.js
```diff
@@ -518,6 +517,13 @@ export default class ExpressionParser extends LValParser {
         this.takeDecorators(node);
         return this.parseClass(node, false);
 
+      case tt.hash:
+        if (this.hasPlugin("classPrivateProperties")) {
+          return this.parsePrivateName(true);
```
Qantas94Heavy: Is this a leftover `true` from the removal of the "liberal" parameter?

ast/spec.md
diervo: I believe semantics are different. 
It doesn't make sense to have an Identifier that has its private attribute to true, first because its only available in the scope of a ClassBody, seconds because its meaning of it is different on the spec.

src/parser/expression.js
diervo: Yep will update the PR

ast/spec.md
loganfsmyth: I could have been clearer, my suggestion is

```
interface PrivateName <: Expression, Pattern {
  type: "PrivateName";
  name: Identifier;
}
```
I'm not suggesting a `private: boolean` flag on `Identifier`.

`Identifier` nodes in our spec are currently _every_ case of non-keyword, non-string literals. The same way it's not `name: string` on `FunctionDeclaration` names or `MemberExpression` properties.

diervo: That I like better. @littledan What do you think?

littledan: Well, thinking of unquoted property names or private field names as Identifiers isn't all that intuitive for me--they are more like strings with a prettier syntax. But if this is just an AST and the rest of the system works this way, it seems fine. I agree that these are somewhat analogous to property names, as long as they are marked specially.

diervo:
@loganfsmyth with this we should be good to go! :)

src/types.js
```diff
@@ -62,6 +62,13 @@ export type Identifier = PatternBase & {
   __clone(): Identifier;
 };
 
+export type PrivateName = PatternBase & {
+  type: "PrivateName";
+  name: string;
+
+  __clone(): Identifier;
```
andy-ms: This presumably clones to a `PrivateName` and not an `Identifier`.

hzoo: I think it's only `this.#a`, `#a` and `#a = 1` in the class body? At least from the examples.

diervo: That's correct

Issue #264, opened 2016-12-21
Parsing error with <= in as a computed key


hzoo:
Ref https://github.com/facebookincubator/create-react-app/issues/1286

### Input Code

```js
const map = {
  [age <= 17] : 'Too young' // error on "="
};
```

http://astexplorer.net/#/g2PDLGAM14

xtuc:
* REPL with [latest](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=&code=const%20map%20%3D%20%7B%0A%20%20%5Bage%20%3C%3D%2017%5D%20%3A%20'Too%20young'%20%2F%2F%20error%20on%20%22%3D%22%0A%7D%3B)
* REPL with [react](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=react&code=const%20map%20%3D%20%7B%0A%20%20%5Bage%20%3C%3D%2017%5D%20%3A%20'Too%20young'%20%2F%2F%20error%20on%20%22%3D%22%0A%7D%3B)

chicoxyzzy:
> The bug seems related to the React preset.

Looks like the problem is in Flow syntax plugin

vkurchatkin:
This is the culprit: https://github.com/babel/babylon/commit/f7c1af1c1f986c5b389faba83657d40d1eac31ae

hzoo:
cc @danharper if you have time

danharper:
Sure, will take a look later today!

vkurchatkin:
Also this probably should work as well:

```js
const map = {
  [<Foo/>] : null
};
```

vkurchatkin:
@hzoo @danharper I've made a quick patch, PTAL

hzoo:
https://github.com/babel/babylon/issues/266 Thanks @vkurchatkin!

PR #266, opened 2016-12-22
Don't set inType flag when parsing property names

vkurchatkin:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | https://github.com/babel/babylon/issues/264
| License           | MIT

### DIFF
--- a/src/plugins/jsx/index.js
+++ b/src/plugins/jsx/index.js
@@ -416,6 +416,10 @@ export default function(instance) {
     return function(code) {
       let context = this.curContext();
 
+      if (this.state.inPropertyName) {
+        return inner.call(this, code);
+      }
+
       if (context === tc.j_expr) {
         return this.jsxReadToken();
       }
--- a/src/tokenizer/state.js
+++ b/src/tokenizer/state.js
@@ -16,6 +16,7 @@ export default class State {
       this.inFunction =
       this.inGenerator =
       this.inAsync =
+      this.inPropertyName =
       this.inType =
       this.noAnonFunctionType =
         false;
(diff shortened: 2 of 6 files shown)


test/fixtures/flow/regression/issue-264/actual.js
```diff
@@ -0,0 +1,3 @@
+const map = {
+  [age <= 17] : 'Too young'
```
xtuc: Do we need the original use case? It's not important in the expected AST.

I would suggest something like that:
```js
[1 >= 0]: null
```

src/tokenizer/state.js
```diff
@@ -16,6 +16,7 @@ export default class State {
       this.inFunction =
       this.inGenerator =
       this.inAsync =
+      this.inPropertyName =
```
danez: Can you please also add the flow type, like the other fields?

src/plugins/jsx/index.js
```diff
@@ -416,6 +416,10 @@ export default function(instance) {
     return function(code) {
       let context = this.curContext();
 
+      if (this.state.inPropertyName) {
+        return inner.call(this, code);
+      }
```
danez: We could move this before the `curContext()` call.

src/tokenizer/state.js
xtuc: I did it in 1d5932c27cfc1ff9ecebbeb3e9ccf11f448cdc26

PR #267, opened 2016-12-24
Add parser support for Flow shorthand import type

jeffmo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| License           | MIT

In a coming version of Flow we'll be adding a new short-hand form of `import type` syntax. [Here's the PR to add it to Flow proper with more details in the summary](https://github.com/facebook/flow/pull/2890).

This just adds parsing support to Babylon for the new syntax in the Flow parser plugin. I'll still need to go through and update the Babel Flow-transform omit `"type"` and `"typeof"` specifiers from printed output.


src/plugins/flow.js
```diff
@@ -1190,6 +1190,58 @@ export default function (instance) {
...
+      } else if (firstIdent.name === "typeof") {
+        specifier_type_kind = "typeof";
+      }
+
+      if (this.isContextual("as")) {
+        const as_ident = this.parseIdentifier(true);
+        if (specifier_type_kind !== null && !this.match(tt.name)) {
+          // `import {type as ,` or `import {type as }`
+          specifier.imported = as_ident;
+          specifier.kind = specifier_type_kind;
```
jeffmo: Oops, this should be `s/kind/importKind/`, update coming...

src/plugins/flow.js
```diff
@@ -1190,6 +1190,58 @@ export default function (instance) {
...
   });
 
+  // parse import-type/typeof shorthand
+  instance.extend("parseImportSpecifier", function () {
+    return function (node) {
+      let specifier = this.startNode();
+      const firstIdentLoc = this.state.start;
+      const firstIdent = this.parseIdentifier(true);
+
+      let specifier_type_kind = null;
```
loganfsmyth: Can we stick with camelCase to be consistent with the rest of the codebase?

jeffmo: Oops... Coming right up

jeffmo:
Thanks! Once this is published, I can update the relevant package.json over in https://github.com/babel/babel/pull/5035 to point to this update

PR #274, opened 2017-01-09
Add support for invalid escapes in tagged templates

bakkot:
Per [the stage-3 TC39 proposal](https://github.com/tc39/proposal-template-literal-revision).

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | maybe?
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | ~~no/yes resp.~~ Edit: yes/yes.
| Fixed tickets     | https://github.com/babel/babel/issues/4798 ~~(progress, not fixed)~~
| License           | MIT

In particular, it allows Babylon to parse ``tag`\u{invalid}` `` and similar.

~~This PR is WIP, but I need feedback from maintainers to finish it. In particular:~~ Edit: I believe it's ready to go pending review.

1. ~~This does not yet have tests. On the other hand, it doesn't cause any test failures either, other than changing the message in a few (which I think is actually an improvement). Where would tests go?~~ Edit: now has many tests. Changes a handful of existing test expectations because it now gives slightly more precise error locations and also does not distinguish between the various kinds of illegal escape sequences in templates.
2. ~~This is enabled unconditionally, not behind a plugin. Presumably it should be, but I'm not sure what all that involves.~~
3. I'm not sure if this counts as a breaking change. It does extend the AST format, but [babel doesn't enforce any constraints on the relevant property](https://github.com/babel/babel/blob/39d18679e1e6f50b7141473d9bf404c0e73bb8fd/packages/babel-types/src/definitions/es2015.js#L331-L333).
4. ~~This causes less precise error and error locations for invalid escape sequences in untagged templates: the error is instead for the whole template token, instead of being at location of the invalid sequence.~~ Also, the message is always "invalid escape sequence", instead of specifying whether it's an octal literal (always disallowed), a `\x` or `\u` where no code point was readable, or a `\u{...}` where the code point specified is > 10FFFF.

~~For the last point, I see a few options:~~

Edit: went with option 2, except only for location information. If it's thought desirable I could give more precise messages as well, instead of (as currently) just reporting "invalid escape sequence in template".

1. Give up on providing precise location of invalid escapes in untagged templates.
2. Extend the tokenizer state with a "previous invalid escape sequence" property, containing its location and message. (This is what e.g. v8 does.) This would be analogous to the `containsOctal` and `octalPosition` properties the tokenizer state already has; it could even just make use of these. 
3. Overload the "value" property of template tokens to be either the cooked value (as it is now) or an object containing the location and message of the first invalid escape sequence.
4. Change the "value" property of template tokens to be either `{ valid: true, cooked }` or `{ valid: false, errorLoc, errorMessage }`.
5. Something else?


hzoo:
> This is enabled unconditionally, not behind a plugin. Presumably it should be, but I'm not sure what all that involves.

Yeah if its not stage-4 it can be behind a plugin - can check another proposal as a reference (https://github.com/babel/babylon/pull/163/files). Basically just running `this.hasPlugin(string)`

bakkot:
Updated to put it behind a plugin named `templateInvalidEscapes`.

src/parser/expression.js
```diff
@@ -316,7 +316,7 @@ pp.parseSubscripts = function (base, startPos, startLoc, noCalls) {
     } else if (this.match(tt.backQuote)) {
       let node = this.startNodeAt(startPos, startLoc);
       node.tag = base;
-      node.quasi = this.parseTemplate();
+      node.quasi = this.parseTemplate(true);
```
xtuc: I'm not a fan of passing boolean as argument in a function. I think it reduce readability.

I would suggest creating two *separated* functions: `parseTaggedTemplate` and `parseTemplate`. What do you think about that?

src/tokenizer/index.js
```diff
@@ -591,16 +591,16 @@ export default class Tokenizer {
 
   // Read a string value, interpreting backslash-escapes.
 
-  readCodePoint() {
+  readCodePoint(throwOnInvalid) {
```
xtuc: Throw on invalid is the default behavior. Could you add a default value?

```js
readCodePoint(throwOnInvalid = true)
```

src/tokenizer/index.js
```diff
@@ -613,7 +613,7 @@ export default class Tokenizer {
       if (ch === quote) break;
       if (ch === 92) { // '\'
         out += this.input.slice(chunkStart, this.state.pos);
-        out += this.readEscapedChar(false);
+        out += this.readEscapedChar(false, true);
```
xtuc: Note: this is a good example of what I explained above. Reading line says nothing on what the function actually does.

src/parser/expression.js
bakkot: Do you mean with the code duplicated, or with both just calling some `parseTaggedOrUntaggedTemplate(isTagged)`?

Either way, if the concern is just passing a raw boolean, how about instead
```js
const isTagged = true;
node.quasi = this.parseTemplate(isTagged);
```

src/tokenizer/index.js
bakkot: Sure, done.

src/tokenizer/index.js
bakkot: Sure. I could rewrite as
```js
const inTemplate = false;
const throwOnInvalid = true;
out += this.readEscapedChar(inTemplate, throwOnInvalid);
```
if you'd prefer.

src/parser/expression.js
xtuc: What about:

```js
node.quasi = this.parseTemplate({ isTagged: true });
```

src/tokenizer/index.js
xtuc: Same as above. I would prefer:

```js
out += this.readEscapedChar({ inTemplate: true, throwOnInvalid: true });
```

Are you ok with that?

src/parser/expression.js
bakkot: Could do, but it would be a break from the style in the rest of Babylon (and, technically, slightly slower).

xtuc: Yes that's true. @danez what do you think about that?

bakkot: Come to think, the same applies to using a default parameter for `readCodePoint`.

xtuc: Yes, you are right, this isn't matching the style of the rest of Babylon.

src/tokenizer/index.js
bakkot: Update: undone to better match style.

bakkot:
I've added tests. As far as I'm aware it should be good to merge pending review.

ast/spec.md
```diff
@@ -949,7 +949,7 @@ interface TemplateElement <: Node {
   type: "TemplateElement";
   tail: boolean;
   value: {
-    cooked: string;
+    cooked: string | null;
```
xtuc: You could use the shorthand: `?string`. Maybe less explicit.

src/tokenizer/index.js
bakkot: @xtuc, I've tried to keep the style in this PR in keeping with the rest of babylon. I have no strong preference either way, but it seemed better to keep the style consistent for the moment. If maintainers have a particular style you'd like me to keep to I can go through and revise (and also revise my other PRs, maybe).

hzoo:
Sorry for the delay @bakkot! I know this is Stage 4 now as well

PR #277, opened 2017-01-10
Estree compatibility as plugin

### LINKED ISSUE #278: Output ESTree (via a plugin)
Lots of previous discussions on this:

- https://github.com/babel/babel/issues/2769
- https://github.com/babel/babel/issues/3813#issuecomment-245406624 (where I said we would add an option)
- https://github.com/babel/babel-eslint (already does the conversion from babylon to estree+esprima to make babel compatible with eslint)
- https://github.com/babel/acorn-to-esprima - previous extracted logic.

PR is https://github.com/babel/babylon/pull/277

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | Fixes #278
| License           | MIT

New plugin to enable estree compatible output.

A new test runner was added, which runs all tests which are throwing tests with the estree plugin. All tests that should throw in non-estree should still throw in estree mode.

- [x] Add compare test between babylon-estree and other parser (flow and/or espree) 
- [x] https://github.com/estree/estree/issues/6


fkling:
This is great, thank you @danez! Any idea when this could be released? I'm happy to help getting this out sooner rather than later.

danez:
I'm currently working on getting `ObjectMethod`/`ObjectProperty` back to `Property`.
That should be already everything that is left, as far as I can see right now.
So I don't want to give any estimates, that I can't hold, but I think it should be ready next week.

fkling:
That is great, thank you. I will wait then 😃

DrewML:
This is such a big deal!

danez:
Quick update, it already works as expected (with some minor things missing, see top) but I'm still working on doing cross-testing with flow and espree, which already revealed some issues.

danez:
I did not commit the comparison tests between babylon with estree and espree/flow as there are a lot of mismatches. Most of them are that espree/flow cannot parse the fixture (and yes all experimental stuff excluded). I need to go through this at some point and fix on our end or report to the other parsers.

Issue #282, opened 2017-01-10
[7.0] Remove node 0.10 and 0.12 from travis


danez:
We are going to drop the support for node 0.10 and 0.12 in version 7.0.0 and a branch for 7.0 does already exist here https://github.com/babel/babylon/tree/7.0.

We can now safely remove node 0.10 and 0.12 from .travis.yml. Also the whole custom logic with nvm in this file can be removed. It was added because rollup only supports node 4 and up.
https://github.com/babel/babylon/blob/7.0/.travis.yml#L11-L14
https://github.com/babel/babylon/blob/7.0/.travis.yml#L20-L21

PR welcome :)

chicoxyzzy:
What about Node 5 which is also deprecated?

chicoxyzzy:
I think we can close this issue :)

PR #284, opened 2017-01-11
[7.0] Remove node 0.10, 0.12 and 5 from Travis

chicoxyzzy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | yes
| Spec compliancy?  | no
| Tests added/pass? | no
| Fixed tickets     | #282 
| License           | MIT


chicoxyzzy:
Possibly we also should add [engines](https://docs.npmjs.com/files/package.json#engines) field to package.json

danez:
Thank you. 

Yes good catch with package.json. Can you add this?

package.json
```diff
@@ -11,6 +11,10 @@
     "bin",
     "lib"
   ],
+  "engines" : {
+    "node" : ">=4.2.0 <5.0.0 || >=6.9.0",
+    "npm": ">=2.14.7"
```
danez: I think we want to go with simply doing >4 for node, and we don't require any special npm version, so we also do not need to put a npm line here.

chicoxyzzy: npm 2.14.7 is a version which comes with Node 4 LTS Argon. Node 4 version proir to 4.2 [are not maintained](https://github.com/nodejs/LTS). So I suppose this should be than
```
{
  "node": ">=4.2",
  "npm": ">=2.14.7"
}
```

danez:
Thank you very much.

PR #285, opened 2017-01-11
Fail tests that have expected.json and throws-option

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | no
| Fixed tickets     | 
| License           | MIT

I can fix the now failing tests at some point, but just noticed it during testing and don't have time right now.

If anyone wants to do this, feel free to either add commits here (as maintainer), or copy my commit into another PR. 
(I would probably do it with a one time script or regex to parse the test output for expected.json locations and do rm on them.)


danez:
Removed with 
```
node_modules/.bin/ava -v 2>&1 | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" | ggrep -P -o "([^\s]*?)/actual.js" | sed s/actual\.js/expected.json/ | sed s/^/test\\/fixtures\\// | xargs rm
```

PR #290, opened 2017-01-12
[7.0] Throw on trailing comma with rest-spread

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

This reverts commit 5bac6e8ad99bcbf608f6df30d9942d6269f88fac. #154

Fixes babel/babel#5092


danez:
Seems travis is confused and still tests 0.10 and 0.12 in this PR although base is 7.0

loganfsmyth:
@danez I see that ESLint still adds this comma when you enable `comma-dangle`, so this is likely gonna cause lots of problems for people. 

I wonder if we should add a "plugin" to allow commas here if people opt into it, so upgrading 6.x to 7.x is easier? We should also really fix ESLint.

loganfsmyth:
Nevermind, my test case was wrong :P

PR #300, opened 2017-01-15
Don't parse class properties without initializers when classProperties plugin is disabled, and Flow is enabled

DrewML:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes (technically)
| New feature?      | no
| Deprecations?     | yes
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

On `master`, we'll currently parse the following, even with `classProperties` disabled

```js
class Foo {
    bar;
}
```

This was happening because we defer the check for the plugin until we're inside the body of `pp.parseClassProperty`, so that `Flow` can parse the following:

```js
class Foo {
    bar: string;
}
```

Once inside of `pp.parseClassProperty`, we were only checking for the plugin if an initializer had been provided.

We unfortunately can't just disallow Flow annotations for class properties when when `classProperties` are disabled, because Flow allows you to do:

```js
class Foo {
   bar: string;

   constructor() {
       this.bar = 'bar';
   }
}
```

I marked this as a breaking change, but it's also technically a bug, so unsure if we want to defer to Babylon 7 or not.

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -771,8 +771,13 @@ pp.parseClassBody = function (node) {
 };
 
 pp.parseClassProperty = function (node) {
+  const noPluginMsg = "You can only use Class Properties when the 'classProperties' plugin is enabled.";
+  if (!node.typeAnnotation && !this.hasPlugin("classProperties")) {
+    this.raise(node.start, noPluginMsg);
+  }
+
   if (this.match(tt.eq)) {
-    if (!this.hasPlugin("classProperties")) this.unexpected();
+    if (!this.hasPlugin("classProperties")) this.raise(this.state.start, noPluginMsg);
     this.next();
     node.value = this.parseMaybeAssign();
   } else {
--- a/test/fixtures/experimental/class-properties/with-initializer-and-type-no-plugin/actual.js
+++ b/test/fixtures/experimental/class-properties/with-initializer-and-type-no-plugin/actual.js
@@ -1,0 +1,3 @@
+class Foo {
+  bar: string = 'bizz';
+}
--- a/test/fixtures/experimental/class-properties/with-initializer-and-type-no-plugin/options.json
+++ b/test/fixtures/experimental/class-properties/with-initializer-and-type-no-plugin/options.json
@@ -1,0 +1,4 @@
+{
+  "plugins": ["flow"],
+  "throws": "You can only use Class Properties when the 'classProperties' plugin is enabled. (2:14)"
+}
--- a/test/fixtures/experimental/class-properties/with-initializer-missing-plugin/actual.js
+++ b/test/fixtures/experimental/class-properties/with-initializer-missing-plugin/actual.js
@@ -1,0 +1,3 @@
+class Foo {
+  bar = 'bizz';
+}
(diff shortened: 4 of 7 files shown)


src/parser/statement.js
```diff
@@ -771,8 +771,13 @@ pp.parseClassBody = function (node) {
 };
 
 pp.parseClassProperty = function (node) {
+  const noPluginMsg = "You can only use Class Properties when the 'classProperties' plugin is enabled.";
+  if (!node.typeAnnotation && !this.hasPlugin("classProperties")) {
```
DrewML: I'm not super happy about a flow-specific check leaking into the general parser, but I didn't see a straight-forward way to do this otherwise. Only other way I could see was to just have the flow extension for `parseClassProperty` to invoke this with an optional second prop `hadType`.

src/parser/statement.js
```diff
@@ -771,8 +771,13 @@ pp.parseClassBody = function (node) {
...
+    this.raise(node.start, noPluginMsg);
+  }
+
   if (this.match(tt.eq)) {
-    if (!this.hasPlugin("classProperties")) this.unexpected();
+    if (!this.hasPlugin("classProperties")) this.raise(this.state.start, noPluginMsg);
```
DrewML: Pretty sure it's safe to provide a better message here. We _should_ only get to this point if we're inside of a class and we've found either a `=`, or an identifier with a line break after it.

DrewML: Should note I'm using `this.state.start` rather than `node.start` here, so we're pointing out the error where the `=` is, rather than what could be the start of a type annotation in the scenario that the code being parsed is:

```js
class Foo {
   bar: string = 'bizz';
}
```

src/parser/statement.js
danez: This code will not stay forever anyway, as as soon as class-properties are final, we can remove all this.

src/parser/statement.js
danez: Yes this was the same before, as `this.unexpected()` uses `this.state.start`

src/parser/statement.js
hzoo: We really need to do https://github.com/babel/babylon/pull/178

DrewML:
> This could be potentially breaking, for people relying on this wrong behaviour. Although I'm not sure if that'S a real case, as babel would not transform class properties when they are not active, but flow is and you use class properties in your code

@danez looks like this wouldn't affect most Babel users.

- If Flow is enabled, [the property without the initializer gets stripped out](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact&code=class%20Foo%20%7B%0A%20%20a%3B%0A%7D&experimental=true&loose=true&spec=false&playground=false).

- If you don't have `Flow` enabled, you already get a syntax error.

This seems like it'd only break things consuming the parser directly (like `babel-eslint`), but I can't think of a way it could actually break code being built.

src/parser/statement.js
DrewML: > This code will not stay forever anyway, as as soon as class-properties are final, we can remove all this.

Good point

danez:
I landed this in 7.0 now.

DrewML:
Thanks @danez!

PR #305, opened 2017-01-17
Disallow import type { type a } from ''

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

I saw that flow was failing, but babylon parsed it just fine.

```
import typeof {typeof t} from "foo";
import type {type t} from "foo";
```

@jeffmo Can you confirm that this is correct?


jeffmo:
Yes this invariant is correct, I must've overlooked it. Thanks for the follow-up!

test/fixtures/flow/type-imports/invalid-import-type-shorthand/options.json
```diff
@@ -0,0 +1,3 @@
+{
```
jeffmo: I'm probably just not familiar with how the test runner/fixtures work, but shouldn't this output have 2 throws (one for each line)?

src/plugins/flow.js
```diff
@@ -1223,6 +1223,10 @@ export default function (instance) {
         specifier.local = specifier.imported.__clone();
       }
 
+      if (node.importKind !== "value" && specifier.importKind !== null) {
```
jeffmo: Per this discussion with @loganfsmyth, I guess this should stick to explicitly listing out `"type"` and `"typeof"`

https://github.com/babel/babel/pull/5035#discussion_r93822476

test/fixtures/flow/type-imports/invalid-import-type-shorthand/options.json
danez: You are right, babylon throws at the first error, and never parses the second line. I extracted that in a second test.

src/plugins/flow.js
danez: Changed that in https://github.com/babel/babylon/pull/312

Issue #331, opened 2017-01-27
[Flow] Add support for method boolean on FunctionTypeAnnotation


vjeux:
Right now,

```js
type T = { method: () => void };
type T = { method(): void };
```

produce the exact same AST (and behave the same way). It would be nice to differentiate somehow so that we can print them the way they were written.

The flow parser is going to get a `method` boolean on `FunctionTypeAnnotation` that distinguishes between the two forms.

It would be great to have the same for Babylon so we can keep the printer simple!

Thanks!

cc @hzoo @samwgoldman

samwgoldman:
Yup, `method` property on the object type ast, like we have for object values, sounds good to me.

danharper:
I've got a PR open now for this (#334), adding `method` to `FunctionTypeAnnotation`.

How about for `ObjectTypeCallProperty` which also contains a method-looking `FunctionTypeAnnotation`, e.g. `type T = { (): number }`?

phpnode:
@danharper `ObjectTypeCallProperty` is always a method, I don't think it's necessary there.

PR #334, opened 2017-01-27
Add method property to FunctionTypeAnnotation node

danharper:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #331
| License           | MIT

For distinguishing the two Flow object type fn styles. See #331, will align with Flow.

```js
type T = { a: () => void };
type T = { a(): void };
```

* [ ] Double check Flow's adding `method` to the `FunctionTypeAnnotation` and not the `ObjectTypeAnnotation`
* [x] ~~How about `ObjectTypeCallProperty` which also contains a method-looking `FunctionTypeAnnotation`? `type T = { (): number }` - it's currently `false`~~

### DIFF
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -316,10 +316,11 @@ pp.flowParseObjectTypeIndexer = function (node, isStatic, variance) {
   return this.finishNode(node, "ObjectTypeIndexer");
 };
 
-pp.flowParseObjectTypeMethodish = function (node) {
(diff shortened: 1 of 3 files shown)


hzoo:

danez:

Thanks for your effort.

PR #335, opened 2017-01-29
Switch to `jest`.

wtgtybhertgeghgtwtg:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes/no
| Fixed tickets     | 
| License           | MIT

Switches from `ava` to `jest` for https://github.com/babel/babel/issues/5197.
Not ready to merge; had to disable `babel-plugin-babel-helpers` for the tests to run.
I'm almost certain I missed something else.


wtgtybhertgeghgtwtg:
Travis, that was it.  This is breaking for dev.

wtgtybhertgeghgtwtg:
Okay, so, `rollup` throws if `modules` is `true`, but `jest` throws if `modules` is false.  And they both run on `BABEL_ENV=test`.

xtuc:
Nice work :+1:. I noticed that Jest ran only one test suite (`test/index.js`). Could we tell Jest to consider each fixtures as a test suite to get a better output?

I was thinking, you could use it programmatically in `test/utils/runFixtureTest.js`. What do you think?

danez:
Van you check that the tests are not slower. Because right now we run ~3500 tests in 7seconds and it shouldn't get any slower.

wtgtybhertgeghgtwtg:
> Could we tell Jest to consider each fixtures (or directory) as a test suite to get a better output?

Not quite.  You can do
```
describe(fixtureName, () => {
  // Run tests here
});
```
but they'll still all be considered one suite.  This will also cause them to run on the same thread, meaning it will be slower than on `ava` (about 10-13 seconds instead of 7).  I'll submit an issue to `jest` to see if this can be reworked.  Something like
```
suite(fixtureName, () => {
  // Run tests here as their own suite
});
```

wtgtybhertgeghgtwtg:
> `rollup` throws if `modules` is `true`, but `jest` throws if `modules` is false.  And they both run on `BABEL_ENV=test`.
Can I get some help on this?  Is there a decision for this?

danez:
Imho jest does not need to do anything with babel, as the source gets precompiled with rollup or babel anyway. So babel-jest is not needed here.

danez:
I just tried it locally and it works fine, but jest currently by default enables babel which is not what their docs say. So I opened a ticket with them: https://github.com/facebook/jest/issues/2848

wtgtybhertgeghgtwtg:
Conflicts resolved, but you'll probably wanna run `yarn upgrade` afterwards, since I did the `yarn.lock` by hand.

danez:
Code coverage is not working anymore. I have no Idea why, but can't get it working. Maybe jest has some more magic that interferes with our coverage reports.

wtgtybhertgeghgtwtg:
`jest` has code coverage built in instead of using `nyc`.  I completely forgot to fix that, sorry.

danez:
The build in only works when jest also does the instrumentation, but we are not doing that, we instrument the code already when we build the rollup bundle.

wtgtybhertgeghgtwtg:
The only thing I can think of is that `jest` doesn't create a `.nyc_output` folder.  I'm not sure if that'd be something they support or would want to support, either.

hzoo:
Yeah, I think jest it's own code coverage but we are using babel-plugin-instanbul already? Now I don't know if it's necessary to switch then?

Issue #340, opened 2017-02-01
allow break/continue outside loop/switch


thejameskyle:
It would be nice to have the following options to be used in babel-template:

- `allowBreakOutsideLoop`
- `allowContinueOutsideLoop`
- `allowBreakOutsideSwitch`

thejameskyle:
Moving to #341

PR #341, opened 2017-02-01
Add allowUnsyntacticBreakContinue option

thejameskyle:
Moving from #340

I went with this option instead because there's a whole validation process with both types of statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,9 @@ Significant diversions are expected to occur in the future such as streaming, EB
 
 - **allowSuperOutsideMethod** TODO
 
+- **allowUnsyntacticBreakContinue**: When enabled, break and continue
+  statements are not validated.
+
 - **sourceType**: Indicate the mode the code should be parsed in. Can be
   either `"script"` or `"module"`.
 
--- a/src/options.js
+++ b/src/options.js
@@ -7,6 +7,7 @@ export const defaultOptions: {
   allowReturnOutsideFunction: boolean,
   allowImportExportEverywhere: boolean,
   allowSuperOutsideMethod: boolean,
+  allowUnsyntacticBreakContinue: boolean,
   plugins: Array<string>,
   strictMode: any
 } = {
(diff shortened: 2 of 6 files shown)


thejameskyle:
We should enable this by default in babel-template when added

loganfsmyth:
For the context of `babel-templates` it feels like this API might be nicer as providing a default context, like "parse as if you're already inside a function" and "parse as if you're already inside a loop" since that focuses on the standard spec behavior, rather than having the flag focus on breaking standard parsing rules.

Thoughts?

thejameskyle:
> parse as if...

are you saying either or or both?

xtuc:
To be sure to understand well, could you please provide a use case for this?

This will be valid, right?
```js
a:
for(;;) {
  break b;
}
```

danez:
I don't like all this options to make the parser parse invalid stuff, but I totally see the usecase in babel-template. Maybe instead of adding more and more options for this, I would suggest creating one option to disable all this checks aka `disableCodeChecks` or similar. That way we would at least have only one option instead of multiple ones for babel-template.

Even cooler would be to separate this logic (in this case) that checks for valid break/continue into an own method `pp.checkBreakContinue()` and then provide a plugin that overrides it and makes it succeed always. That way we wouldn't have to do all the `if (option) {}` logic inside the actual parsing code.
Not sure if that works in all the cases we currently cover with options, but would be worth a try.
Thoughts?

loganfsmyth:
> are you saying either or or both?

Assuming I'm understanding your question, I think either-or?

`babel-templates` already takes options that pass through to Babylon, so my proposal would be an attempt to generalize cases where parsing as if it had a given parent, which is kind of like what you seem to be looking to solve.


What I had in mind would be like

```
options = {
  context: {
    // Same as "allowReturnOutsideFunction"
    function: true,

    // Same as "allowSuperOutsideMethod"
    method: true,
  },
};
```

then your case we could do

```
// Parse, allowing break and continue
loop: true
```

We just landed `.parseExpression` but you could imagine that being

```
// Parse, assuming the parser is already in an expression
expression: true
```

as a replacement for that.

Similarly, we probably want
```
// Parse "yield" without explicit wrapper "function*(){ ... }"
generator: true,

// Parse "await" without explicit wrapper "async function(){ ... }"
async: true,
```

danez:

Thanks for your effort.

Issue #348, opened 2017-02-07
[7.0] Change ForAwaitStatement node type to match ESTree


hzoo:
Ref https://github.com/estree/estree/pull/138

> Change to be `ForOfStatement with property "await": true`

Original PR: https://github.com/babel/babylon/pull/17

---
- [ ] Comment below you are going to do this (for others to know)
- [ ] Read [CONTRIBUTING.md](https://github.com/babel/babylon/blob/master/CONTRIBUTING.md)
- [ ] Check you are on 7.0 branch with `git checkout 7.0` and then make a branch off of that to PR
- [ ] modify spec.md doc (table of contents + entry) https://github.com/babel/babylon/blob/master/ast/spec.md#forawaitstatement
- [ ] modify node type https://github.com/babel/babylon/blob/898c4a7623787795c35b57ce119b4fd1a29d1766/src/parser/statement.js#L536
- [ ] fix test fixtures

aweary:
@hzoo mind if I give this one a shot? 😃

hzoo:
sure go ahead 🙂, and join our slack if you haven't already

hzoo:
Done in https://github.com/babel/babylon/issues/349

PR #349, opened 2017-02-08
Remove ForAwaitStatement, add await flag to ForOfStatement

aweary:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | https://github.com/babel/babylon/issues/348
| License           | MIT

Removes the `ForAwaitStatement` node type and adds an `await` flag to the existing `ForOfStatement` node!

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -530,12 +530,11 @@ pp.parseFor = function (node, init) {
 // same from parser's perspective.
 
 pp.parseForIn = function (node, init, forAwait) {
-  let type;
+  const type = this.match(tt._in) ? "ForInStatement" : "ForOfStatement";
   if (forAwait) {
+    node.await = true;
     this.eatContextual("of");
-    type = "ForAwaitStatement";
   } else {
-    type = this.match(tt._in) ? "ForInStatement" : "ForOfStatement";
     this.next();
   }
   node.left = init;
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -532,11 +532,11 @@ pp.parseFor = function (node, init) {
 pp.parseForIn = function (node, init, forAwait) {
   const type = this.match(tt._in) ? "ForInStatement" : "ForOfStatement";
   if (forAwait) {
-    node.await = true;
     this.eatContextual("of");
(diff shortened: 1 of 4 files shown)


hzoo:
Oops forgot to add the extra steps about changing to 7.0 branch (did it)

src/parser/statement.js
```diff
@@ -530,12 +530,11 @@ pp.parseFor = function (node, init) {
 // same from parser's perspective.
 
 pp.parseForIn = function (node, init, forAwait) {
-  let type;
+  const type = this.match(tt._in) ? "ForInStatement" : "ForOfStatement";
   if (forAwait) {
+    node.await = true;
     this.eatContextual("of");
-    type = "ForAwaitStatement";
   } else {
```
xtuc: Should we set the `node.await` to `false` there? According to the changes in the spec I would expect a `boolean`.

Since `undefined` is falsy this isn't an issue.

hzoo: Do we do this elsewhere for other flags (false/true)? I think we might

xtuc: I also think so. Don't have much context to answer you. Maybe @danez can.

aweary: Using `Function` as a reference, the `generator` and `async` flags are explicitly set to `false` when they don't apply, so it seems like it should do the same here.

aweary:
@xtuc @hzoo I updated so the `await` flag should always be set on `ForOfStatement` nodes.

xtuc:
I updated the Babel 7 upgrade guide accordingly in babel/babel.github.io@dd2586aaacd45a0157fdff763f24f1240659fc61

danez:
We need a follow-up story in babel-types, babel-generator and the transforms using `ForAwaitStatement` to account for this changes.

hzoo:
why I wish babylon was in the monorepo again

danez:
I created a ticket babel/babel#5286

PR #351, opened 2017-02-10
Fix parsing of class properties

bakkot:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes (kind of? some edge cases now give different ASTs)
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #116, #157, #158
| License           | MIT

This completes the work begun by @motiz88 in #158.

It fixes a couple of issues:
- ASI is no longer performed after the `get` or `a` below:
```js
class A {
  get
    x(){}

  a
    (){}
}
```
- Class properties separated by semicolons may appear on one line: `class A { x; y; }` works now.
- Class properties may now be named `static`.
- EDIT: Static properties named `prototype` and non-static properties named `constructor` are now forbidden.


danez:
Thanks, what are the changes from #158 ?

bakkot:
@danez, the description is (I think) everything that's changed. It's based off the cases found in #158 but doesn't end up using the same approach - as we found in #158, getting the current code to be correct is tricky; the obvious changes lead to other issues. I think this approach is going to be easier to maintain, and is to my eye more obviously correct (well, and is also actually correct, unlike the current code).

Sorry about the repetition; it's difficult to avoid if we want to preserve useful error messages. I'll take another look to see if I can cut it down. Happy to take suggestions there too.

FWIW, this approach is very similar to the approach I took when implementing this in V8. I'm reasonably confident in it except the parts relating to flow, which I have less experience with. Unfortunately the flow parser itself has some bugs here.

Issue #33, opened 2016-05-14
Decorators on computed class method names


leonid-bauxy:
I propose that support for decorators on computed properties is added to `babylon`.

It was suggested in `babel-eslint` repo, that I open an issue here.
Related issue: https://github.com/babel/babel-eslint/issues/309.

This doesn't work right now:

``` js
  @autobind
  [method](arg) {
      // do something
  }
```

hzoo:
This is handled then https://github.com/babel/babylon/pull/587 and will be in the next beta release which should be 7.0.0-beta.15 (soon)

PR #353, opened 2017-02-10
Update decorator grammar to match current spec

bakkot:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #13, #33, #231, #232
| License           | MIT

The current [decorators proposal](https://github.com/tc39/proposal-decorators) only allows very specific productions in a decorator, not arbitrary AssignmentExpressions. This updates the grammar to match the spec.

Likely to break some code.

Ping @DrewML re: #236.


danez:
Thanks, the current implementation of decorators is really old, and we probably do not want to change it anymore, but instead will create a new plugin for the new spec.

So it might take some time, till we know how to continue with all the decorator changes.

bakkot:
@danez Makes sense. Gonna go ahead and close this, but feel free to take the code and/or tests when that revision happens.

PR #360, opened 2017-02-15
Fix generator-method-with-computed-name test

rattrayalex:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes (test fix)
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | not sure
| Tests added/pass? | yes
| License           | MIT

This test was previously checking that a block does not parse as an object, though the intent appeared to be to validate that you can have `[yield iter]` be the key of an object. 

I'm not sure if this is compliant with the spec, or if it makes any sense. If this is indeed supposed to raise an error, I'd be happy to try to submit a patch fixing the parser.


danez:
All parsers seem to parse this correctly so I assume this is correct.

PR #363, opened 2017-02-18
Add back range property

hzoo:
We'll want to make this an option?


danez:
Although it doesn't hurt to have it added all the time, but yes we could make it an option.

hzoo:
Less options the better though, dono

PR #377, opened 2017-02-23
Don't consume async when checking for async func decl

existentialism:
As reported in https://github.com/babel/babel/issues/5366.

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -835,6 +835,13 @@ pp.parseExport = function (node) {
     let needsSemi = false;
     if (this.eat(tt._function)) {
       expr = this.parseFunction(expr, true, false, false, true);
+    } else if (
+      this.isContextual("async") &&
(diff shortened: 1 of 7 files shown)


src/parser/statement.js
```diff
@@ -835,6 +835,12 @@ pp.parseExport = function (node) {
     let needsSemi = false;
     if (this.eat(tt._function)) {
       expr = this.parseFunction(expr, true, false, false, true);
+    } else if (
```
existentialism: @danez not sure if there's a cleaner way to do this check w/o consuming?

danez:
Isn't this the exact same thing as the just reverted commit? This would reintroduce the problem. The problem is that babel can't handle the new AST.

danez:
Oh I see this is about arrow functions. Before we merge this we need to fix babel first to understand the AST.

hzoo:
If necessary we can backport to 6.x

PR #381, opened 2017-02-24
Enable debugging a single fixture using TEST_GREP

aaronang:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | no
| Fixed tickets     | 
| License           | MIT

I was trying to debug a single fixture to get a better understanding of #370. However, there was no option that allowed me to do this? Therefore, I implemented `TEST_GREP` such that one can run a certain fixture by providing a (partial) file path.

```bash
$ TEST_GREP="flow/tuples" ./node_modules/.bin/inspect node_modules/ava/profile.js test/index.js
```

Running this command will open a Chrome DevTools instance.

I feel like the current implementation is really hacky, so feedback would be appreciated.


package.json
```diff
@@ -34,6 +34,7 @@
     "eslint-config-babel": "^6.0.0",
     "eslint-plugin-flowtype": "^2.20.0",
     "flow-bin": "^0.38.0",
+    "inspect-process": "^0.4.2",
```
loganfsmyth: I don't know how others feel, but I usually see personal utils like this as good to install globally, since many people may test things in their own way.

aaronang: That makes sense. I will also update it in babel/babel#5365.

test/utils/runFixtureTests.js
```diff
@@ -3,10 +3,14 @@ var getFixtures = require("babel-helper-fixtures").multiple;
 
 exports.runFixtureTests = function runFixtureTests(fixturesPath, parseFunction) {
   var fixtures = getFixtures(fixturesPath);
+  var testGrep = process.env.TEST_GREP || "";
 
   Object.keys(fixtures).forEach(function (name) {
     fixtures[name].forEach(function (testSuite) {
       testSuite.tests.forEach(function (task) {
+        if (task.actual.loc.indexOf(testGrep) === -1) {
```
loganfsmyth: Since this defaults to `""`, won't this always be `0`? "thing".indexOf("") === 0`

aaronang: Yes, so if someone does not define `TEST_GREP` it will just run all test cases.

loganfsmyth: Ugh, this is what I get for late evening code review :P

package.json
existentialism: Agreed

danez:
This is already possible, but not documented. Simply add `"only":true` to the options.json or create the file with this content if not exist yet.

hzoo:
Ah yeah `only: true` does work and should be documented

aaronang:
Okay, I will close this issue and document it today 👍

hzoo:
I'm ok with a command if it makes it easier, but I guess only works alright (unless we support a grep to run multiple, but you probably only want to look at one at a time anyway)

aaronang:
I was using `TEST_GREP` to run only one test case, and sometimes it ran more than one because it matched multiple test cases, which was a bit annoying. I think `only: true` should be fine 😄

PR #384, opened 2017-02-26
[7.0] Change RestProperty/SpreadProperty to RestElement/SpreadElement

hzoo:
- [x] need to fix a failing test

  estree-throws › es2015/uncategorised/288
  /Users/hzhu/dev/babylon/test/utils/runFixtureTests.js:92

   91:   if (opts.throws) {
   92:     throw new Error("Expected error message: " + opts.throws + ". But parsing succeeded.");
   93:   } else {

  Error: /Users/hzhu/dev/babylon/test/fixtures/es2015/uncategorised/288/actual.js: Expected error message: Invalid left-hand side in assignment expression (1:1). But parsing succeeded.

Odd that `[...a, b] = c` is parsing as a SpreadElement when it should be a RestElement?


src/parser/lval.js
```diff
@@ -34,8 +34,8 @@ pp.toAssignable = function (node, isBinding, contextDescription) {
         this.toAssignable(node.value, isBinding, contextDescription);
         break;
 
-      case "SpreadProperty":
-        node.type = "RestProperty";
+      case "SpreadElement":
```
danez: SpreadElement should only be converted to RestElement if isBinding === true and otherwise fail.

Previously we were not converting `SpreadElement` at all, but now we need to in some cases (the cases that were previously SpreadProperty)

hzoo: If you have time to finish this up, go ahead 👍

PR #393, opened 2017-03-02
Enable type-checking in src

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     | None
| License           | MIT

This change enables type-checking in all `src` files (via `all=true` in `.flowconfig`).

Two big changes were needed:

* Prototype assignment (`Parser.prototype.foo = function () { ... }`) doesn't get a type-checked `this`. Changed these to `Parser.mixin(class extends Parser { ... }`, and did the actual prototype assignment in `mixin`.
* Similarly, instead of using `parser.extend`, a plugin is now a function `(superClass: Class<Parser>): Class<Parser> => class extends superClass { ... }`, and `parser.usePlugin` does the actual extending (and setting of `super`).

Other than those, I tried as much as possible to keep the code as it is, and just add type annotations and comments.

Notes:

* For mixed-in methods used outside of the mixin class, they must be declared in `parserTypes.js` or they are invisible to the outside.
* This disables eslint's `indent` rule so that it can get a readable diff. Indenting could be fixed in another PR.
* This declares `[key: string]: any` on `Node`, since often a node will be downcast to another node type and writing all those downcasts would be a pain.
* `estree.js` contains lots of `$FlowIgnore` since we're changing the node to a different type.
* Added `throw` in front of `this.raise(...)` and `this.unexpected(...)` so that control-flow analysis can know about these.


.eslintrc
```diff
@@ -2,5 +2,8 @@
   "extends": "babel",
   "env": {
     "node": true
+  },
+  "rules": {
+    "indent": 0
```
danez: Why did you disable this rule? Where there problems with it?

src/options.js
```diff
@@ -1,17 +1,19 @@
...
-} = {
+export type Options = {
+  sourceType: "script" | "module";
+  sourceFilename?: string;
+  startLine: number;
+  allowReturnOutsideFunction: boolean;
+  allowImportExportEverywhere: boolean;
+  allowSuperOutsideMethod: boolean;
+  plugins: $ReadOnlyArray<string>;
+  strictMode: any;
```
danez: strictMode can be boolean or null

.eslintrc
andy-ms: If I indent things correctly, this PR will have an unreadable diff because the indent of most lines will change. Don't know a good solution for that.

danez: Huh? Wasn't it properly indented before?

src/plugins/flow.js
```diff
@@ -995,77 +987,64 @@ export default function (instance) {
...
-
-  instance.extend("toAssignable", function (inner) {
-    return function (node, isBinding, contextDescription) {
+    }
+
+    // don't lex any token as a jsx one inside a flow type
+    jsx_readToken() {
+      // $FlowFixMe This method doesn't override anything!
+      if (!this.state.inType) return super.jsx_readToken();
+    }
```
danez: Looking at the history this was never working, so can you please remove this?

src/plugins/estree.js
```diff
@@ -116,40 +114,32 @@ export default function (instance) {
...
-    return function (node, ...args) {
-      inner.call(this, node, ...args);
+    parseBlockBody(node: N.BlockStatementLike, ...args): void {
+      super.parseBlockBody(node, ...args);
 
-      node.directives.reverse().forEach((directive) => {
-        node.body.unshift(this.directiveToStmt(directive));
-      });
+      const directiveStatements = node.directives.map((d) => this.directiveToStmt(d));
+      node.body = directiveStatements.concat(node.body);
```
danez: reverse() got missing.

andy-ms: `reverse` + `forEach` + `unshift` = `concat`. :relaxed:

src/types.js
```diff
@@ -0,0 +1,676 @@
...
+  end: number;
+  loc: SourceLocation;
+};
+
+export interface NodeBase {
+  start: number;
+  end: number;
+  loc: SourceLocation;
+  range: [number, number];
+  leadingComments?: ?Array<Comment>;
```
vkurchatkin: Flow has a [bug](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAdgFwKYBOUAhgMZ5gAaYA3qmGDHsQCbYDmAwnALY964AzgC4wAfgCCBAsQCeAHm58BOAHwBuVAF90pOFkE4wAD1HUAvLXqNmbLF178hogNoBdbeqA) with maybe-typed properties in interfaces. It's better to avoid interfaces completely unless they only contain methods

src/parser/comments.js
```diff
@@ -25,20 +25,20 @@
...
-function last(stack) {
+function last<T>(stack: $ReadOnlyArray<T>): T {
   return stack[stack.length - 1];
 }
 
-const pp = Parser.prototype;
-
-pp.addComment = function (comment) {
+Parser.mixin(class extends Parser {
+addComment(comment: N.Comment): void {
```
vkurchatkin: You can omit annotations if you don't export class or function

src/parser/node.js
```diff
@@ -1,48 +1,68 @@
...
+  leadingComments: ?Array<N.Comment>;
+  trailingComments: ?Array<N.Comment>;
+  innerComments: ?Array<N.Comment>;
+  extra: { [key: string]: any };
+
   constructor(parser?: Parser, pos?: number, loc?: Position) {
+    // $FlowIgnore
     this.type = "";
+    // $FlowNonNull
     this.start = pos;
```
vkurchatkin: If it's always present, then why `pos` is optional?

src/parser/statement.js
```diff
@@ -238,6 +244,7 @@ pp.parseForStatement = function (node) {
     if (forAwait) {
       this.unexpected();
     }
+    // $FlowCast node: N.ForStatement
```
vkurchatkin: It's better to do this explicitly with `any`:  `((node: any): N.ForStatement)`

src/parser/comments.js
andy-ms: Some of them could be omitted, but I've tested around and some of them can't without breaking compilation. I also think it's nice to have the type annotations locally so I don't have to look them up in `parserTypes.js`. The code worked fine without *any* type annotations, the point was to make it easier to read.

src/parser/node.js
andy-ms: `new Node;` is called in one place, but I just added a `$FlowFixMe` comment at that place and no longer need hacks here. Thanks!

src/parser/statement.js
andy-ms: Done

src/types.js
andy-ms: This is implemented by a class, so I think we need the interface?

src/parser/expression.js
```diff
@@ -352,31 +355,31 @@ pp.parseCallExpressionArguments = function (close, possibleAsyncArrow) {
   }
 
   return elts;
-};
+}
 
-pp.shouldParseAsyncArrow = function () {
+shouldParseAsyncArrow(): boolean {
```
Kovensky: Does flow not infer this, or do you want to assert that it returns a boolean?

src/parser/expression.js
```diff
@@ -454,12 +457,13 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
         return this.finishNode(node, "DoExpression");
       }
 
-    case tt.regexp:
+    case tt.regexp: {
       const value = this.state.value;
-      node = this.parseLiteral(value.value, "RegExpLiteral");
+      const node = this.parseLiteral(value.value, "RegExpLiteral");
```
Kovensky: A similar change could be made in most of the switch cases, but probably best to not make the diff larger

src/parser/index.js
```diff
@@ -1,11 +1,41 @@
...
+   */
+  static mixin(mixedIn: Class<Parser>): void {
+    const proto: any = Parser.prototype;
+    const mixedProto: any = mixedIn.prototype;
+    for (const key of Object.getOwnPropertyNames(mixedProto)) {
+      if (key === "constructor")
+        continue;
+      if (proto[key])
+        throw new Error(`Can't mixin '${key}', already defined`);
+      proto[key] = mixedProto[key];
```
Kovensky: The copied keys will be enumerable, but they are not when defined in the `class` expression.

On the other hand, before the change to `class` expression, they were enumerable anyway, so it might not matter.

src/parser/index.js
```diff
@@ -57,20 +83,35 @@ export default class Parser extends Tokenizer {
...
+    for (const key of Object.getOwnPropertyNames(extenderProto)) {
+      if (key === "constructor") {
+        continue;
+      }
+      if (!self[key]) {
+        throw new Error(`Can't extend '${key}', doesn't exist`);
+      }
+
+      superProto[key] = self[key];
+      self[key] = extenderProto[key];
```
Kovensky: Same thing in these two lines here w.r.t. enumerability.

src/parser/expression.js
andy-ms: For a function declared in `parserTypes.ts`, the return type *should* be carried along from there. But for a function declared only locally, you may get no (or confusing) errors. I think it's better to just consistently declare types. E.g., remove the return type from `parseCallExpressionArguments`, add `elts.push(0);` somewhere, and you won't get any errors.

src/parser/index.js
andy-ms: I don't have an opinion on this.

src/parser/expression.js
andy-ms: Whoops, must have been trying something and forgot to change it back.

hzoo:

hzoo:
Hey @andy-ms! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/213598556)
* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/213598557)

andy-ms:
It looks like the test failure is in babel somewhere -- did this PR change it or was the change in babel?

hzoo:
Yeah a different pr we need to fix that I merged

hzoo:
Hey @andy-ms! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/216022710)
* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/216022711)

andy-ms:
Hm, still failing.
```
TypeError: [BABEL] /home/travis/build/babel/babylon/node_modules/istanbul-lib-instrument/dist/index.js: Plugin 0 specified in "/home/travis/build/babel/babylon/.babelrc.env.test" was expected to return a function but returned "undefined" (While processing preset: "/home/travis/build/babel/babylon/node_modules/babel-preset-stage-0/lib/index.js") (While processing preset: "/home/travis/build/babel/babylon/node_modules/babel-preset-es2015/lib/index.js")
```

existentialism:
Yeah, looks like an unrelated (to this PR) issue with the latest b7 alpha, will take a look after lunch.

andy-ms:
Seems to have been fixed.

hzoo:
Hey @andy-ms! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/219372620)
* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/219372621)

andy-ms:
Again, test failures are probably due to some change in babel?

danez:
Besides it being a massive diff I do not like they idea of restructuring code just to please flow. It might have been easier to start small and just enable flow file by file and PR them separately.
This current approach with the inheriting classes and then the separate file with the type definitions seems rather complicated and unusual to me, and we maybe should think about a better structure of the Parser instance that would also work with flow.

Ohh and we should keep an eye on performance, as a drastic change like this might also have performance impacts.

loganfsmyth:
I'll second performance. The `super` change could certainly introduce performance changes since it has to look up the prototype chain instead of calling a reference to the function it already has.

That said, I don't have a problem with the structure. I think having the node types in their own file is inevitable since they are used all over the place, and the class extension + mixin approach is a little weird, not doesn't seem too bad to me.

andy-ms:
Bad news: It looks like with plugins applied, this branch does make parsing a bit slower.
Good news: We can make it a lot faster!

Currently, plugins work by writing functions directly to the *instance*.
This branch modifies plugins to be functions from a superclass to a new subclass.
But it then takes the methods off of that subclass and directly assigns them...

I have modified this to instead just create a class, and never directly assign to an instance.
These classes can be reused across multiple `parse` calls, and node seems to have an easier time optimizing them.

Here are my measurements parsing `jquery` with the `flow` and `jsx` plugins applied:

| Name                           | Time |
| ------------------------------ | ---- |
| master, no plugins             | 7.86 8.07 8.02 8.14 7.84 |
| type-check, no plugins         | 7.93 8.05 7.80 7.97 7.96 |
| master with plugins            | 22.0 21.2 21.6 21.0 |
| type-check with plugins        | 25.3 24.1 23.0 22.9 |
| type-check with cached plugins | 10.6 10.5 10.3 10.0 |

It looks like the existence of instance functions was making it run significantly slower, but moving that all into a prototype brought the time back down.

Shall I make a pull request doing just this?

hzoo:
Hey @andy-ms! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/219813778)
* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/219813779)

andy-ms:
Closing since this has been broken into many separate PRs.

PR #402, opened 2017-03-07
Throw error if new.target is used outside of a function

### LINKED ISSUE #400: new.target can only be used in functions
### Input Code

```js
new.target
```

### Expected Behavior
Error: "`new.target` can only be used inside functions"

### Current Behavior
parses just fine

new.target is only valid if it is inside a function  `function () { new.target }` but not outside of functions.

flow, acorn, espree, esprima fail correctly

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | Fixes #400
| License           | MIT


src/parser/expression.js
```diff
@@ -542,6 +542,10 @@ pp.parseMetaProperty = function (node, meta, propertyName) {
     this.raise(node.property.start, `The only valid meta property for new is ${meta.name}.${propertyName}`);
   }
 
+  if (!this.state.inFunction) {
+    this.raise(node.property.start, "new.target can only be used in functions");
+  }
```
danez: Could this be triggered also by `function.send`? If so I would rather do this check before/after calling this method.

existentialism: Good catch, updated!

(Also, TIL about `function.sent` lol)

PR #405, opened 2017-03-10
Fix push-pop logic in flow

### LINKED ISSUE #404: Un-popped push
In the `src/plugins/flow.js` [override](https://github.com/babel/babylon/blob/7a6d495704d4d86189ef13195ad3115cca8f5bbc/src/plugins/flow.js#L1375) of `parseMaybeAssign`, there is a call to `this.state.context.push()`, followed by a block of code, followed by a call to `this.state.context.pop()`.
But there is an early return inside that block of code, meaning that the push may not be popped. This looks incorrect, but naively fixing it by adding a `pop()` before the early `return` breaks some tests.

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | 
| New feature?      | 
| Deprecations?     | 
| Spec compliancy?  | 
| Tests added/pass? | yes
| Fixed tickets     | Fixes #404
| License           | MIT

@andy-ms I think that should be correct now, and tests seem to still work.


hzoo:

hzoo:
Hey @danez! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/213754746)
* [Node vlts/* - linux](https://travis-ci.org/babel/babylon/jobs/213754747)

PR #408, opened 2017-03-12
Improves error message when super is called outside of constructor

arshabh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| License           | MIT

This PR improves the error message thrown when `super()` is used outside of the `constructor` method. 

Many developers coming from Java background use the name of the class itself as the constructor method name. Also sometimes the developers make a typo like `constructr` or `contructor` and wonder what is the issue. This message would help them in understanding the issue and fix it faster.

Opened a PR on babel [here](https://github.com/babel/babel/pull/5450) to make sure that the tests pass there as well when this PR is merged

### DIFF
--- a/src/parser/expression.js
+++ b/src/parser/expression.js
@@ -392,7 +392,7 @@ pp.parseExprAtom = function (refShorthandDefaultPos) {
         this.unexpected();
       }
       if (this.match(tt.parenL) && this.state.inMethod !== "constructor" && !this.options.allowSuperOutsideMethod) {
-        this.raise(node.start, "super() outside of class constructor");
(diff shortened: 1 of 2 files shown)


hzoo:
"and has no typos", doesn't seem necessary to me, I'l merge without that part

existentialism:
What about:

> Found `super()` inside method `construtor`, it is only valid inside a class constructor.

hzoo:

PR #409, opened 2017-03-13
[7.0] Replace non-existent RestPattern in the ast spec with RestElement

Jamesernator:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | N/A
| Fixed tickets     | 
| License           | MIT

Basically the spec defines ObjectPattern.properties to incorrectly be `[ AssignmentProperty | RestPattern ]` which refers to a non-existent node RestPattern, as Babylon returns a RestElement in such cases I replaced it in the spec.


hzoo:
Thanks!!

PR #415, opened 2017-03-17
[7.0] Moved value field in spec from ObjectMember to ObjectProperty as ObjectMethod's don't have it

Jamesernator:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | N/A
| License           | MIT

`ObjectMember` was defined as having a value field but actually only `ObjectProperty` has this field but `ObjectMethod` extends `ObjectMember` so it shouldn't have it there, so I removed it off `ObjectMember` and put it on `ObjectProperty` instead.


hzoo:
Nice catch!

Kind of thinking we should just link to ESTree again and then add the nodes we have that are different (Babel 6 nodes)

PR #435, opened 2017-03-27
Reuse plugin "extend" function definitions

zertosh:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | no
| Fixed tickets     | 
| License           | MIT

The functions passed to `instance.extend` do not themselves need to be re-defined every time you create a new instance. I know this PR creates history noisy, so no hard feelings if you don't want to merge it. Also, I'd be more than happy to backport it to 6.x.

To best see the changes, use `?w=1`: https://github.com/babel/babylon/pull/435/files?w=1


danez:
Do you think this would be better with #393 ?

zertosh:
@danez looks like no. In #393, they're defining new "extension plugin" classes every time you create a new parser instance. Seems like a trade-off between perf and types :/

danez:
I think this might have been fixed now with #459 as the actual class now gets cached. 
Anyway this seems to conflict heavily with the other PR, so we should probably close this, but if you have further ideas how to improve babylon open another PR.
Thanks for looking into this.

zertosh:
Ah the class caching approach is so much better, thanks @danez!

PR #436, opened 2017-03-27
Spec fix as BindExpression only have a single object/callee not an Array

Jamesernator:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | N/A
| Fixed tickets     |
| License           | MIT

Basically BindExpression in the spec wraps the properties in an Array, but Babylon just returns a single item (which is what makes sense anyway).


hzoo:
good catch, really weird that I did that.. https://github.com/babel/babel/pull/3295

PR #438, opened 2017-03-28
Fix #437: only prohibit 'export type from "module" ' when flow is enabled

### LINKED ISSUE #437: export type from "module" is a SyntaxError
### Input Code
Basically this is a SyntaxError:
```js
export type from "./some-module.js"
```

[example](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=true&presets=stage-1&targets=&browsers=&builtIns=false&code=%0Aexport%20type%20from%20%22.%2Ffoo.js%22&experimental=true&loose=false&spec=false&playground=true&stage=0)

Using Babylon plugin: ['exportExtensions'] but it affects Babel too using the babel plugin 'transform-export-extensions'.

### Current Behavior
Currently this results in a SyntaxError, I'm fairly certain it's because of flow parsing which uses `export type ...`.

### Your Environment

| software         | version
| ---------------- | -------
| Babylon          |  6.16.1
| node             | v7.7.3
| npm              | 4.1.2
| Operating System | Ubuntu

bakkot:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  |no
| New feature?      |no
| Deprecations?     |no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #437
| License           | MIT

Moves the logic prohibiting `export type from "module"` to the flow plugin.


hzoo:
Hey @bakkot! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

loganfsmyth:
I don't think this should be an error even with Flow. Ideally we'd look to see if it is followed by `from` otherwise take the Flow parsing branch?

bakkot:
@loganfsmyth, no strong opinion on what the behavior should be with flow. This is the intersection of a stage-1 extension and flow; it's probably fine to throw in sufficiently weird corner cases, especially until flow decides what it wants to do.

hzoo:
Hey @bakkot! It looks like one or more of your builds have failed. I've copied the relevant info below to save you some time.

loganfsmyth:
Ah right, I didn't connect that it was the experimental proposal. You have a  fair point on it being an edge case. Do you have a sense for how easy it would be to handle like I mentioned? If it's a pain I'm fine leaving it for now.

bakkot:
It wouldn't be all that hard - it's not quite like you say, because `from` is a legal type name in flow, so you need two-token lookahead to tell between flow's `export type from = foo` and ES6's `export type from ""` (or `export type, {asdf} from ""`).

But looking at it again, it's a weird enough case that I think it *should* be an error for now, unless and until the official flow parser (or grammar - is there an official flow grammar?) decides what it wants to do in this case. If flow eventually decides *not* to support it, then it would be a breaking change for babylon to remove support, whereas adding support could be done in a patch release.

Jamesernator:
Although an edge case this is something I actually hit in actual code, specially I made a tiny library for making modules automatically from folders ([here](https://www.npmjs.com/package/folder-module)) which used the `export <name> from "./folder-with-module/<name>.js"` for its exporting, for now though I can just switch it to 

```js
import foo from "./<name>.js"
export foo
```

as currently `transform-export-extensions` doesn't transform `export { <local> as <exported> } from "<module>"` [currently](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=true&presets=stage-1&targets=&browsers=&builtIns=false&code=export%20%7Bfoo%20as%20bar%7D%20from%20%22.%2Ffoobar.js%22&experimental=true&loose=false&spec=false&playground=true&stage=0) but I'll put in an issue on Babel for that.

bakkot:
@Jamesernator This PR makes your code work unless you have Flow enabled, in which case it probably shouldn't work (any more than `export class from "m"` should).

For `export { a as b } from "m" `, that's not an *extension* to the export syntax; that's part of es2015. I don't think it's something the export-extensions plugin ought to transform.

Jamesernator:
I didn't know that that was actually part of es2015, good to know thanks.

bakkot:
Force-pushed to trigger CI again after #442; now passes.

Issue #413, opened 2017-03-15
Tighten up parsing of Flow TypeCasts


DrewML:
### Input Code

```js
state.set('foo': bar)
```

### Babylon/Babel Configuration

Flow syntax plugin enabled

### Expected Behavior
Should be a syntax error

### Current Behavior
We're currently parsing `bar` as a TypeCastExpression w/ a value of `foo`. Flow requires that you wrap a TypeCastExpression in parens, and we're allowing the parens for the CallExpression to be used right now, which the Flow parser does _not_ allow.

Context: https://twitter.com/schep_/status/841741581589860355

dschep:
Thanks for pointing out this is actually even a bug if you ARE using flow. Not a solution I expected to my problem 😄

PR #446, opened 2017-04-02
Parses flow type cast expression with parenthesis in call expression correctly

chikara-chan:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #413 
| License           | MIT

By the way, now the async arrow expression is not parsed from call expression any more. Instead, check if `async` is a keyword or function name there. Then decides to parse call expression arguments or function params.


src/parser/expression.js
```diff
@@ -298,22 +323,23 @@ pp.parseSubscripts = function (base, startPos, startLoc, noCalls) {
       this.expect(tt.bracketR);
       base = this.finishNode(node, "MemberExpression");
     } else if (!noCalls && this.match(tt.parenL)) {
-      const possibleAsync = this.state.potentialArrowAt === base.start && base.type === "Identifier" && base.name === "async" && !this.canInsertSemicolon();
+      const old = this.state.clone(true);
```
danez: I'm a little bit worried about this and doing it whenever we encounter a parenL. clone() is very expensive and we have never done lookaheads of this extend before (besides one token with `this.lookahead()`).

wcjohnson:
It is almost certainly possible to fix this without using parser lookahead. See https://github.com/wcjohnson/babylon-lightscript/commit/322842abe7b9fd0ece731eaa2238edeaaa315876 for an outline of an approach.

Basic outline:
- parse from within parenLists as `TypeCastish`
- only convert to `TypeCastExpression` from within `parseParenAndDistinguishExpression`
- lvalue conversion converts `TypeCastish` into the correct annotations
- if `TypeCastish` is seen anywhere else, raise an unexpected type cast error

danez:

Thanks for your effort.

Issue #440, opened 2017-03-28
Discussion: Should we have sourceType: "auto"


loganfsmyth:
In Babel 6, we've hardcoded everything to be `sourceType: "module"` by default, which has been a bit of a problem because people often run Babel on arbitrary code that they don't own, and sometimes that code is written assuming it will be treated as a script. This leads to complaints about `this` being replaced at the top level, among other things.

Would people be open to adding an `auto` sourceType option that would use the unambiguous grammar proposal? https://github.com/bmeck/UnambiguousJavaScriptGrammar

The longterm effect I'd expect from this for users is that users who have normal CommonJS modules won't automatically get `use strict` at the top of their files anymore, unless they opt into that plugin, which I think it a very reasonable thing to do.

The other side of it is that I want to start locking down the behavior of modules specifically, so usage of `module.exports` or `require` alongside `import` would be an error, which I believe is also the case in Webpack now? We'll want a flag so people can opt out of the behavior when switching from Babel 6 to 7, though.

Jamesernator:
Yes that'd be awesome.

danez:
In webpack you can use require alongside with import, which I think is a common pattern if you want to lazy-require stuff or but requires in `if (NODE_ENV==='development' )` so the whole require+the required dependencies get stripped on production. import() is not an alternative as webpack treats import() as splitpoint by default, though this can be turned of by changing the parser configuration in webpack.

loganfsmyth:
@danez Assuming I'm understanding the planned behavior in Node right, `require` and the others like `module.exports` and such will not exist inside an ES module. That means if you want to use `require` you'd have to be write a CommonJS module, and then import that from your ES6 module.

I'm still on the fence about adding that restriction to Babel as default in 7.x, may just be an opt-in flag. Webpack 2 does not implement the restriction of using `require`, but they _do_ disallow using `exports` alongside ES6 module syntax, so that we should probably do at least.

Either way I do think it's beneficial for Babel to default to "auto", because it means that people don't need to explicitly set `sourceType`. Most people don't know about it, and Babel doesn't currently expose a way to toggle it based on extension or something, so you can't mix and match in a codebase. With `auto`, if you run Babel on a file without `import` or `export`, we can stop automatically adding `use strict` and we won't automatically add `.__esModule` and we won't replace `this` with `undefined`, which are all problems people tend to hit unexpectedly because most people don't know about the grammar split.

jdalton:
[...]

PR #449, opened 2017-04-05
Support sourceType: "unambiguous" parsing

loganfsmyth:
| Q                 | A
| ----------------- | ---
| Bug fix?          | N
| Breaking change?  | N
| New feature?      | Y
| Deprecations?     | 
| Spec compliancy?  | 
| Tests added/pass? | N
| Fixed tickets     | #440
| License           | MIT

This code is ugly so definitely consider a WIP, but I'm posting for feedback before I go any further. This is the same logic used by Webpack to parse files and will result in files being treated as modules if they contain at least one import/export statement.

My proposed usage of this in Babel itself would be to change Babel's current default `sourceType: "module"` to `sourceType: path.extname(opts.filename || "") === ".mjs" ? "module" : "script"` so that files that Babel can begin toggling module transpilation directly based on `sourceType`, and can throw errors if import statements are inserted into a CommonJS module for instance. Similarly, this means we can stop adding `use strict` at the top of CommonJS files automatically and stop rewriting `this` to `undefined`.

With this, users can opt into `sourceType: unambiguous` in their `.babelrc` to tell Babel to essentially guess what type of file it is. This same logic is already implemented in Webpack and a new other loaders.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,7 +18,27 @@ plugins.jsx = jsxPlugin;
 plugins.typescript = typescriptPlugin;
 
 export function parse(input: string, options?: Options): File {
-  return getParser(options, input).parse();
+  if (options && options.sourceType === "unambiguous") {
+    options = Object.assign({}, options);
+    try {
+      options.sourceType = "module";
+      const ast = getParser(options, input).parse();
+
+      // Rather than try to parse as a script first, we opt to parse as a module and convert back
+      // to a script where possible to avoid having to do a full re-parse of the input content.
+      if (!hasModuleSyntax(ast)) ast.program.sourceType = "script";
+      return ast;
+    } catch (moduleError) {
+      try {
+        options.sourceType = "script";
+        return getParser(options, input).parse();
+      } catch (scriptError) {}
+
+      throw moduleError;
+    }
+  } else {
+    return getParser(options, input).parse();
+  }
 }
 
 export function parseExpression(input: string, options?: Options): Expression {
@@ -90,4 +110,17 @@ function getParserClass(
     parserClassCache[key] = cls;
   }
   return cls;
+}
+
+function hasModuleSyntax(ast) {
+  return ast.program.body.some(
+    child =>
+      (child.type === "ImportDeclaration" &&
+        (!child.importKind || child.importKind === "value")) ||
+      (child.type === "ExportNamedDeclaration" &&
+        (!child.exportKind || child.exportKind === "value")) ||
+      (child.type === "ExportAllDeclaration" &&
+        (!child.exportKind || child.exportKind === "value")) ||
+      child.type === "ExportDefaultDeclaration",
+  );
 }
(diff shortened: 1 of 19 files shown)


src/index.js
```diff
@@ -19,7 +19,27 @@ plugins.flow = flowPlugin;
...
-  return new Parser(options, input).parse();
+  if (options && options.sourceType === "unambiguous") {
+    options = Object.assign({}, options);
+    try {
+      options.sourceType = "module";
+      const ast = new Parser(options, input).parse();
+
+      // Rather than try to parse as a script first, we opt to parse as a module and convert back
+      // to a script where possible to avoid having to do a full re-parse of the input content.
+      if (!hasModuleSyntax(ast)) ast.program.sourceType = "script";
```
Kovensky: The "correct logic" would be to parse as script first and module later if that fails, but true, it's a lot more likely to get code that parses as both (or module only) as input. Not a lot of people still have code that _only_ parses in sloppy mode.

danez:
Will node do the same now? 
`If extension is mjs then assume module otherwise autodetect?`
Did they finally agree on something?

Kovensky:
Nope! This is still a best guess. The only thing that's certain is that _if_ `mjs` happens that it'll always be parsed as a module.

src/index.js
```diff
@@ -32,3 +52,12 @@ export function parseExpression(input, options) {
 
 
 export { tokTypes };
+
+function hasModuleSyntax(ast) {
+  return ast.program.body.some((child) =>
+    child.type === "ImportDeclaration" ||
+    child.type === "ExportDeclaration" ||
+    child.type === "ExportAllDeclaration" ||
+    child.type === "ExportDefaultDeclaration"
```
loganfsmyth: You still get an AwaitExpression either way, I think that part of the spec is more to make like easier for parsers if they want.

In this example, if you use `await` as an identifier it'll throw and reparse as a script successfully, which I think is good?

src/index.js
loganfsmyth: Yeah, I think it's as long as the output result is the same, the potential better average performance makes me want to go this route.

loganfsmyth:
I think at this point Node will end up with `.mjs` for modules and `.js` for CommonJS. That'll be a slow transition for the community though, so I think it's reasonable to force `module` for `.mjs` and use this for `.js` for a while at least. It'll also keep us more consistent with Webpack in the short term and allow the community to migrate over on their own time.

src/index.js
jdalton: Is it possible to warn if ambiguous code is detected?

src/index.js
jdalton: For perf I had the acorn folks dig into proofing out [an optimized top level parse](https://github.com/RReverser/esmod/blob/master/index.js). In reify we also do a source sniff for `import` or `export` and if it fails we skip it.

danez:

Thanks for your effort.

PR #454, opened 2017-04-06
Refresh property ordering in baselines

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| License           | MIT

Done by running `scripts/rmExpected.js`, then `yarn test`.
We currently don't test for property ordering, meaning that deleting and recreating a test can have a diff. Would adding a test for ordering make a good PR?


danez:
Well they are all json-objects, so ordering does not really matter I would say?

Btw instead of the rm-script one could simply do `rm -rf test/fixtures/**/expected.js`

danez:
Maybe we can use some snapshot library and do snapshot testing, means also ordering would count. The only thing I do not know is if these snapshot libraries provide a good way to show what changed in a massive stringfied json.

I merge this for now.

PR #479, opened 2017-04-21
Ensure non pattern shorthand props are checked for reserved words

### LINKED ISSUE #477: `{this}` should be a SyntaxError
Original issue submitted by @Ginden in https://github.com/babel/babel/issues/5652

Currently Babel parses `{this}` object literal.

It's syntax error in Chrome and Firefox.

The same applies to other keywords:

```
({const, if, this})
```

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes/yes
| Fixed tickets     | Fixes #477
| License           | MIT


danez:
Does this still work with this PR:
```
({ const: 1, this: 2 });
```

existentialism:
Yeah, if we handle it in a [diff path if we detect a colon](https://github.com/babel/babylon/blob/master/src/parser/expression.js#L874).

danez:
Nice.

PR #480, opened 2017-04-21
Add type declarations for AST nodes

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     | 
| License           | MIT

This simply creates the type definition for each AST node kind.
It doesn't yet use the types in any files. However, I have tested these out in #393 and expect to be able to use these with few modifications.


danez:
👍

PR #459, opened 2017-04-09
Convert each plugin to a function from a class to an overriding class

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | tests pass
| Fixed tickets     | 
| License           | MIT

Instead of assigning functions to a parser instance, this creates a class with overridden methods.
The class is a function of the plugins (only "jsx", "flow", and "estree"), and this function is cached.

Based on my earlier comment: https://github.com/babel/babylon/pull/393#issuecomment-292623114
Like #393, this avoids fixing the indent inside each plugin, in order to have a smaller diff.

This will improve performance because we aren't assigning functions directly to the instance any more.
I ran the [perf script](https://gist.github.com/andy-ms/138e270e510620945df1499181c05892) again on my home machine just to be sure:

| branch | time |
| --- | --- |
| `master` | 26.8 26.8 27.1 26.0 26.0 |
| `plugin_class` | 12.8 13.0 13.1 13.3 12.7 |

PR #481, opened 2017-04-21
Convert each file with parser methods to a class in an inheritance chain

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     | 
| License           | MIT

As mentioned in #393, providing a type for `this` is difficult if direct prototype assignment is used. This PR converts each part of the single `Parser` class to a class of its own in an inheritance chain.

This also introduces a new `BaseParser` class with parts split off from `Parser` so that they could be at the top.

The inheritance chain ordering that I went with, from top to bottom, is:

* `BaseParser`
* `CommentsParser`
* `LocationParser`
* `Tokenizer`
* `UtilsParser`
* `NodeUtils`
* `LvalParser`
* `ExpressionParser`
* `StatementParser`
* `Parser`

There is no need to put plugin helpers in this chain, as those should be moved into the plugin itself (#482).

I tested with the [perf script](https://github.com/danez/babylon_performance) and this doesn't seem to have any impact.

Remember to view the diff with `?w=1`: https://github.com/babel/babylon/pull/481/files?w=1 (thanks @STRML https://github.com/babel/babylon/pull/459#issuecomment-295418785)


src/parser/node.js
```diff
@@ -34,43 +34,40 @@ class Node {
...
+  finishNode(node, type) {
+    return this.finishNodeAt(node, type, this.state.lastTokEnd, this.state.lastTokEndLoc);
+  }
 
-pp.finishNodeAt = function (node, type, pos, loc) {
-  return finishNodeAt.call(this, node, type, pos, loc);
-};
+  // Finish node at given position
 
+  finishNodeAt(node, type, pos, loc) {
```
andy-ms: Not sure why this was initially broken into both `pp.finishNodeAt` and `function finishNodeAt`. Seems to work fine this way.

danez: I was also wondering the same. ¯\_(ツ)_/¯

PR #485, opened 2017-04-25
Type-check UtilParser

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     | 
| License           | MIT

Enables type-checking in `util.js` and annotates every method.


hzoo:
Thanks again!

src/parser/util.js
```diff
@@ -32,57 +35,57 @@ export default class UtilParser extends Tokenizer {
...
-  isLineTerminator() {
+  isLineTerminator(): boolean {
     return this.eat(tt.semi) || this.canInsertSemicolon();
   }
 
   // Consume a semicolon, or, failing that, see if we are allowed to
   // pretend that there is a semicolon at this position.
 
-  semicolon() {
+  semicolon(): void {
```
xtuc: Do we need to explicitly write `void` as the returned type?

andy-ms: Without the return type, you should still get errors if there are errors, but they will be more confusing: If you have `function f() { return g(); }` and call `f().toUpperCase()`, and there's an error, is it in `f`, `g`, or `.toUpperCase()`?

xtuc:
I was thinking it will infer the void type since It returns nothing, doesn't it?

andy-ms:
Yes, in the case of 100% correct code everything will be inferred with no problem. But there are also some functions in the code that *should* be void but contain unnecessary return statements -- can't catch that without an annotation.

xtuc:
Ok, thanks

PR #488, opened 2017-04-25
Type-check ExpressionParser

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     |
| License           | MIT

Enables type-checking in the file and annotates all methods.
Requires a few forward-declarations for `statement.js`. `parseClass` might deserve its own file; same for decorator-related methods. But trying to just do type annotations in this PR.


andy-ms:
Codecov is having problems here; don't know where the coverage was lost.
screenshot

PR #491, opened 2017-04-26
Type-check utils

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     | 
| License           | MIT

Enables type-checking in these files and adds type annotations.


src/util/location.js
```diff
@@ -1,18 +1,32 @@
+// @flow
+
 import { lineBreakG } from "./whitespace";
 
+export type Pos = {
+  start: number;
```
hzoo: this actually used anywhere?

andy-ms: I used it as the type for `refShorthandDefaultPos`.

PR #492, opened 2017-04-26
Type-check State

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     |
| License           | MIT

Enables type-checking in the file and adds annotations.
Enabling type-checking here also caused errors in some other files, so fixed those too.


src/tokenizer/index.js
```diff
@@ -626,6 +626,7 @@ export default class Tokenizer extends LocationParser {
       code = this.readHexChar(this.input.indexOf("}", this.state.pos) - this.state.pos, throwOnInvalid);
       ++this.state.pos;
       if (code === null) {
+        // $FlowFixMe (is this always non-null?)
```
hzoo: maybe @bakkot can help out with this?

PR #495, opened 2017-04-26
Type-check flow plugin

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     |
| License           | MIT

Enables type-checking in the file and adds type annotations.
This isn't a very strict check since the spec for the node types isn't done.


src/plugins/flow.js
```diff
@@ -1018,8 +1035,8 @@ export default (superClass) => class extends superClass {
     }
   }
 
-  // declares, interfaces and type aliases
-  parseExpressionStatement(node, expr) {
+  // declares, interfaces and tNype aliases
```
existentialism: Nit: typo

src/plugins/flow.js
```diff
@@ -252,21 +258,24 @@ export default (superClass) => class extends superClass {
...
           // flow does not support the ExportNamedDeclaration
+          // $FlowIgnore
           node.type = "ExportDeclaration";
+          // $FlowFixMe
           node.default = false;
           delete node.exportKind;
         }
 
+        // $FlowIgnore
         node.type = "Declare" + node.type;
```
hzoo: Well stuff like this isn't fun

andy-ms: Type systems don't handle objects changing type very well. There are similar problems in the estree plugin.

hzoo: Yeah it's just not a good way to handle the types, dynamically adding the strings

Issue #500, opened 2017-04-29
Feature Request: An open plugin system


fahad19:
Right now, we are limited to using a set number of plugins that resides in the babylon repository. If there was a way to develop plugins externally, and still use them with Babylon, it would allow others to contribute more.

### Example API

An example could be:

```js
var babylon = require('babylon');
var myCustomPlugin = require('babylon-plugin-custom');

babylon.parse(code, {
  plugins: [
    'jsx',
    'flow',
    myCustomPlugin
  ]
});
```

### Related issues:

* https://github.com/babel/babel/issues/4187

I stumbled upon this because I wanted to write a new Babel plugin that parses a new kind of syntax, but, of course, it was throwing errors before I could do anything with the original source.

hzoo:
Hey @fahad19! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

PR #501, opened 2017-05-01
Export plugins

urish:
This will allow users of the module to define third-party plugins

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | pass
| Fixed tickets     | #500
| License           | MIT

This will allow users of the module to define third-party plugins:

```javascript
    var babylon = require('babylon')

    babylon.plugins.customPlugin = function(parser) {
         ...
    };
    
    babylon.parse('someExpression()', { plugins: [ 'customPlugin'] });
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,7 +35,7 @@ export function parseExpression(input: string, options?: Options): Expression {
   return parser.getExpression();
 }
 
-
+export { plugins };
 export { tokTypes };
 
 function getParser(options: ?Options, input: string): Parser {


fahad19:
💖

loganfsmyth:
Duplicate of https://github.com/babel/babylon/pull/11, https://github.com/babel/babel/issues/4245, and https://github.com/babel/babel/issues/1351, among others I'm sure.

The PR isn't just asking to expose some object, it's asking us to maintain a whole official API for parser plugins to use, which isn't something we're willing to commit to currently, as it's not clear how to make that API effective, and it would limit out ability to refactor and optimize the codebase.

We do allow people to fork `babylon` currently, and use `{ parserOpts: {parser: require("custom-parser")}}` in the options if you want to experiment with something as a custom syntax, but that is the limit of it currently.

urish:
Actually, for my purpose - no API for plugins is needed. If you merge it as-is, without any commitment from your side, I will be super-happy

loganfsmyth:
Your example code takes `parser` as an argument. That is an API. We could change the method names on that instance at any time. If we exposed this, we'd break any custom plugins any time we did that.

urish:
True, but I'm (and perhaps other?) happy to live with exported plugins object despite the fact it might even be removed at some point, if you decide to do so.

It is still so much better than [hacks like this](https://github.com/lukehorvat/babby/blob/b925e8151c6d5f58abe893cb9680221416bfaa4b/lib/babylon.js#L6) people are currently doing.

loganfsmyth:
Currently we don't endorse having plugins that affect syntax. 

> It is still so much better than hacks like this people are currently doing.

"better" here is relative though. If we expose `plugins` as a mutable object, it will encourage people to hack them in anyway. The fact that it is a gross pain in the ass right now is a good thing, because it makes it obvious that it isn't something we support people doing.

> I'm happy to live with exported plugins object despite the fact it might even be removed at some point, if you decide to do so.

But you aren't the person people will be yelling at. If you hack something together, publish it, people start using it, and then we break it because we didn't even know about it and it used a hack to add it in there, average users will have no idea, they'll just see that their Babel-based project is now broken, and we'll be the ones stuck in the position of either leaving users broken, or then having to support the hack long-term.

urish:
Got it. So do you currently recommend on forking babylon, making whatever changes and then publishing to NPM so it can be consumed by babel? Or is there a better way to go?

hzoo:
Yeah that's the recommendation: added to the readme now https://github.com/babel/babylon#will-babylon-support-a-plugin-system

Thanks for the PR though.

PR #509, opened 2017-05-04
Better error message for anonymous class declarations

rattrayalex:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | --
| License           | MIT

Given code like:
```js
class {}
```
or 
```
class extends Bar {}
```
Instead of `Unexpected token (1:6)`, get `Class name required (1:6)`

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -804,7 +804,7 @@ export default class StatementParser extends ExpressionParser {
       if (optionalId || !isStatement) {
         node.id = null;
       } else {
-        this.unexpected();
+        this.unexpected(null, "A class name is required");
       }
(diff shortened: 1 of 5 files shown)


rattrayalex:
Cool - I also like "A class name is required". Thoughts/suggestions from
anyone else?

I'll probably leave as is if nobody else has an opinion if that's cool

rattrayalex:
Awesome, thanks guys! Updated to `A class name is required`

Issue #320, opened 2017-01-22
Typescript Plugin


hzoo:
Ref https://twitter.com/left_pad/status/820054499389804544

This would help tooling for other things like ESLint, prettier, and users that use both Babel and Typescript (and using either < stage 3 syntax or want to use Babel to compile). And maybe other things in the future?

Now I'm not sure anyone on the team is able to dedicate time to create or maintain the plugin (there's plenty of stuff to work on already and all as volunteers) but we can help out with questions (testing, how it works, etc). I think it would work the same as how we maintain the flow plugin. The Flow team submits PRs for upcoming features/fixes and we help to review the changes according to our style/codebase.

A new proposal requires a few changes: https://github.com/babel/babel/wiki/Adding-a-new-Proposal-to-Babel. Basically a change to the parser, the code generator (for consistency although maybe not necessary), the AST spec doc, the 
AST helper functions, and to the plugin that removes the types.

We are determining what to do about spec changes/major bumps via https://github.com/babel/babylon/issues/275. Basically a change that isn't a major bump would be a different plugin so we can just release a minor of the parser rather than a major bump.

The flow team is available to talk to us via slack/github and now we also have a #typescript room in our slack so can also discuss with the TS team and others that want to help out with this effort. (Signup at http://slack.babeljs.io/, the room url after you've signed in is https://babeljs.slack.com/messages/typescript/).

> I'm going to edit this section later about the basics of implementing this but just wanted to get an issue started.

- Babylon is acorn, so the way you can do plugins is by extending certain parser methods
- https://github.com/babel/babylon/tree/master/src/plugins for flow/jsx
  - [flow extending](https://github.com/babel/babylon/blob/b6c3b5aa8319582f695e7944503612e39aa08261/src/plugins/flow.js#L825-L827)
  - [jsx extending](https://github.com/babel/babylon/blob/b6c3b5aa8319582f695e7944503612e39aa08261/src/plugins/jsx/index.js#L402-L403)

---

- [ ] find and extend the correct hooks for TS (probably similar to flow for most, need a few new ones)
- [ ] create all the helper parser methods `pp.methodHere`
- [ ] add to plugins https://github.com/babel/babylon/blob/b6c3b5aa8319582f695e7944503612e39aa08261/src/index.js#L14-L17
- [ ] import tests from TS?
- [ ] update readme

JamesHenry:
[...]

PR #523, opened 2017-05-12
TypeScript parser

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | Tests added, and pass
| Fixed tickets     | #320
| License           | MIT

This is a WIP, so please don't merge yet. The AST spec is likely to change. I am just looking to get comments on the implementation.
For those interested in just the AST spec, just look at the changes to `src/types.js`. I am tracking differences with typescript-eslint-parser [here](https://github.com/JamesHenry/tsep-babylon-test/issues).

CC @stephen @JamesHenry @DanielRosenwasser

Notes:

* It shares a lot in common with the flow plugin, but for now I am just using identical code (and marking it as such). In the future they should probably share.

* Test files should really have a `.ts` extension, but that would require editing `babel-helper-fixtures`. Shall I make a pull request there?


hzoo:
> Test files should really have a .ts extension, but that would require editing babel-helper-fixtures. Shall I make a pull request there?

I think that should be covered by https://github.com/babel/babel/pull/5511 but I didn't think we really wanted it earlier?

hzoo:
I know it's easier to modify the code in src/expression etc, but is there a way for us to make the parser plugin like in estree/flow/jsx? https://github.com/babel/babylon/tree/master/src/plugins.

The concern is how we are going to do versioning? In flow I guess everything is a minor/patch though.

src/parser/expression.js
```diff
@@ -956,16 +1034,17 @@ export default class ExpressionParser extends LValParser {
...
     const oldInMethod = this.state.inMethod;
     this.state.inMethod = node.kind || true;
     this.initFunction(node, isAsync);
     this.expect(tt.parenL);
-    node.params = this.parseBindingList(tt.parenR);
+    const allowModifiers = isConstructor; // For TypeScript parameter properties
+    node.params = this.parseBindingList(tt.parenR, /* allowEmpty */ false, allowModifiers);
     node.generator = !!isGenerator;
-    this.parseFunctionBody(node);
+    const missingBody = this.maybeParseFunctionBody(node, /* allowExpression */ false, /* allowMissingBody */ type === "ClassMethod");
```
xtuc: I think we could pass an object here, like:

```js
maybeParseFunctionBody({ allowExpression: false, allowMissingBody: type === "ClassMethod"});
```

Looks better than comments to me.

src/parser/expression.js
```diff
@@ -1080,29 +1181,43 @@ export default class ExpressionParser extends LValParser {
...
+    return name;
   }
 
   checkReservedWord(word: string, startLoc: number, checkKeywords: boolean, isBinding: boolean): void {
+    if (this.hasPlugin("typescript")) {
+      // Don't bother checking for TypeScript code.
+      // Strict mode words may be allowed as in `declare namespace N { const static: number; }`.
+      // And we have a type checker anyway, so don't bother having the parser do it.
+      return;
+    }
```
xtuc: What about a better integration with TypeScript here? Early grammar errors are great for online editor (like Babel's REPL). Babylon has that kind of logic for JavaScript.

andy-ms: We would have to know of whether we're in a declaration context or not. And knowing that requires knowing the current file extension. `export let package: number;` is a perfectly legal `.d.ts` file.

andy-ms:
> I know it's easier to modify the code in src/expression etc, but is there a way for us to make the parser plugin like in estree/flow/jsx?
Does that mean you would want more changes like 733943a?

src/parser/expression.js
nicolo-ribaudo: > And knowing that requires knowing the current file extension.

We could also add a `"sourceType": "declaration"` which treats files as `.d.ts`

andy-ms:
@nicolo-ribaudo Done https://github.com/babel/babylon/pull/523/commits/bee125d51ced2d80b12b2cc111c9d1f31c456bbd

src/parser/statement.js
```diff
@@ -632,8 +632,14 @@ export default class StatementParser extends ExpressionParser {
...
   parseFunctionParams(node: N.NormalFunction): void {
+    node.params = this.doParseFunctionParams();
+  }
+
+  doParseFunctionParams(): $ReadOnlyArray<N.Pattern> {
     this.expect(tt.parenL);
-    node.params = this.parseBindingList(tt.parenR);
+    // $FlowIgnore (won't be TSParameterProperty)
+    const x: $ReadOnlyArray<N.Pattern> = this.parseBindingList(tt.parenR);
+    return x;
```
nicolo-ribaudo: Why are both `parseFunctionParams` and `doParseFunctionParams` needed?

andy-ms: I needed that specific functionality from within the typescript plugin.
But it looks like it's simpler if I just call `super.parseFunctionParams` to access that instead.

src/parser/expression.js
```diff
@@ -294,99 +294,121 @@ export default class ExpressionParser extends LValParser {
...
         node.computed = true;
+        node.optional = true;
         this.expect(tt.bracketR);
-        base = this.finishNode(node, "MemberExpression");
-      } else if (!noCalls && this.match(tt.parenL)) {
-        const possibleAsync = this.state.potentialArrowAt === base.start && base.type === "Identifier" && base.name === "async" && !this.canInsertSemicolon();
-        this.next();
+        return this.finishNode(node, "MemberExpression");
+      } else if (this.eat(tt.parenL)) {
+        const possibleAsync = this.state.potentialArrowAt === base.start &&
```
nicolo-ribaudo: This looks like a perfect place to use the new `#atPossibleAsync` method

andy-ms: :+1:

src/parser/expression.js
```diff
@@ -1072,8 +1105,27 @@ export default class ExpressionParser extends LValParser {
     return false;
   }
 
+  parseFunctionReturnTypeAndBodyAndFinish(node: N.BodilessFunctionOrMethodBase, type: string, allowExpressionBody?: boolean): void {
```
nicolo-ribaudo: I think this should just be `typescript.js#parseFunctionBodyAndFinish` which overwrites `expression.js#parseFunctionBodyAndFinish`

src/parser/expression.js
```diff
@@ -1163,26 +1215,33 @@ export default class ExpressionParser extends LValParser {
...
+
     if (this.match(tt.name)) {
-      node.name = this.state.value;
+      name = this.state.value;
     } else if (this.state.type.keyword) {
-      node.name = this.state.type.keyword;
+      name = this.state.type.keyword;
     } else {
-      this.unexpected();
+      throw this.unexpected();
```
nicolo-ribaudo: `throw` is not needed -- `#unexpected()` already throws

andy-ms: Flow will complain without the `throw` since it's not clear that control flow can't reach the end without assigning `name`.

nicolo-ribaudo: 👍 I didn't think about that

src/parser/expression.js
```diff
@@ -776,6 +799,11 @@ export default class ExpressionParser extends LValParser {
     node.callee = this.parseNoCallExpr();
     const optional = this.eat(tt.questionDot);
 
+    if (this.isRelational("<")) {
+      const typeArguments = this.parseTypeArguments();
```
nicolo-ribaudo: What if "<" is just a relational operator? Like in this case:
```js
function A() { }
A.prototype.valueOf = () => 2;
var T = 3;

new A < T;
```
I see that in `typescript.js#parseSubscript` you considered that edge case when parsing call expressions.
(I have not tested if this actually works or not, but I got confused by this difference in the two methods)

---

Anyway, you can also extract these lines (799-818 or 802-812) to a separate method and overwrite it in `typescript.js`, thus removing the type arguments check from `expression.js`

hzoo:
Can review the non-test files with https://github.com/babel/babylon/compare/master...andy-ms:ts-wip-no-test?expand=1

src/parser/lval.js
```diff
@@ -30,6 +30,7 @@ export default class LValParser extends NodeUtils {
         case "ObjectPattern":
         case "ArrayPattern":
         case "AssignmentPattern":
+        case "TSParameterProperty":
```
danez: Why not override and do this in the plugin?

src/parser/lval.js
```diff
@@ -190,17 +194,27 @@ export default class LValParser extends NodeUtils {
...
+    this.parseAssignableListItemTypes(left);
+    const elt = this.parseMaybeDefault(left.start, left.loc.start, left);
+    if (decorators.length) {
+      left.decorators = decorators;
+    }
+    return elt;
+  }
+
+  // Overridden in typescript.js
+  parseParameterModifiers(): { accessibility: ?Accessibility, readonly: boolean } {
```
existentialism: Is this still needed? Couldn't see where it was used (or overridden)

src/plugins/flow.js
```diff
@@ -1341,6 +1340,12 @@ export default (superClass: Class<Parser>): Class<Parser> => class extends super
 
   parseAssignableListItemTypes(param: N.Pattern): N.Pattern {
     if (this.eat(tt.question)) {
+      if (param.type !== "Identifier") {
```
existentialism: Is this expected behavior in flow? Using the fixture from the matching typescript test in astexplorer seems to produce a [funky result](https://astexplorer.net/#/gist/b55806bfc15230ce5df73f09335bdaa5/bdcd344d7137134d11afa84271176aa43d87bea7).

If it is, we should probably add a matching flow test fixture.

andy-ms: Flow itself fails for your example with `Unexpected token ?`. I'll add a test.

hzoo:
@danez @stephen @nicolo-ribaudo any more thoughts?

src/parser/expression.js
```diff
@@ -786,6 +795,8 @@ export default class ExpressionParser extends LValParser {
     node.callee = this.parseNoCallExpr();
     const optional = this.eat(tt.questionDot);
 
+    this.parseNewTypeArguments(node);
```
nicolo-ribaudo: I didn't notice this before: is it possible to implement new type arguments only in the typescript plugin?

Something like
```js
parseNew() {
    const node = super.parseNew();
    if (/* node is NewExpression */ && !/* node has args parentheses*/ && this.match("<")) {
       // parse type arguments
       // maybe parse arguments
    }
}
```
---

Anyway, apart from this detail this pr looks good to me! (well, I have to admit I haven't reviwed *all* of it because I don't know typescript very well 😅 )

nicolo-ribaudo:
I have a concern about versioning: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes makes me think that typescript can have breaking changes. Does this PR make babylon *potentially* need a major version bump every time a new ts version is released?

mhegazy:
Typescript never had syntactic breaking changes. Meaning valid syntax that is supported by one version and disallowed by a newer version. All breaking changes are semantic in nature; either new error checks or new behaviors for the type system. All these are outside the scope of Babylon. There are some changes in the resulting code from tsc but these are not relevant here either.
So from parsing perspective we have a guarantee for no breaking changes movinging forward.
Hope that addresses your concern.

nicolo-ribaudo:
Thank you 👍

src/parser/expression.js
andy-ms: I don't think the AST gives us a way to distinguish between `new X` and `new X()`. But I found a different way that works just as well, see the latest commit.

src/plugins/typescript.js
```diff
@@ -1147,16 +1147,16 @@ export default (superClass: Class<Parser>): Class<Parser> => class extends super
...
 
-  parseNewTypeArguments(node: N.NewExpression): void {
-    if (!this.isRelational("<")) { // tsTryParseAndCatch is expensive, so avoid if not necessary.
-      return;
+  parseNewArguments(node: N.NewExpression): void {
+    if (this.isRelational("<")) { // tsTryParseAndCatch is expensive, so avoid if not necessary.
+      // 99% certain this is `new C<T>();`. But may be `new C < T;`, which is also legal.
+      const typeParameters = this.tsTryParseAndCatch(this.tsParseTypeArguments.bind(this));
+      if (typeParameters) {
+        node.typeParameters = typeParameters;
```
nicolo-ribaudo: I think you should check that `this.match(tt.parenL)`, otherwise `new Foo<T>` will be allowed

andy-ms: :+1:

JamesHenry:
Congrats, @andy-ms! Awesome work 😄 

We can keep fine tuning the AST together but it's fantastic to have this merged!

andy-ms:
@hzoo Please notify me when this is published.

hzoo:
yep doing it now!

hzoo:
https://github.com/babel/babylon/releases/tag/v7.0.0-beta.16

yahiko00:
Oh. My. God. <3

PR #525, opened 2017-05-15
Fix ns export when between a default export and exported specifiers.

jdalton:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 127
| License           | MIT

A continuation of #127. I noticed that this will error:
```js
export B, * as A, { C, D } from 'test';
```
The bit that's throwing [it](https://github.com/babel/babylon/blob/7a8b64c0d38e615903b484aa1b3e99ebbd21ddbe/src/parser/statement.js#L822-L888) off is the last specifier `, { C, D }`.

Related to https://github.com/benjamn/reify/pull/162.


loganfsmyth:
My understanding is that this is expected behavior. These proposals essentially add `export` versions of the existing `import` forms, and the ES6 spec for `import` allows one, but not both of `{foo as foo}` or `* as foo`.

Note in the grammar: https://www.ecma-international.org/ecma-262/7.0/#sec-imports

```
ImportClause:
    ImportedDefaultBinding
    NameSpaceImport
    NamedImports
    ImportedDefaultBinding, NameSpaceImport
    ImportedDefaultBinding, NamedImports
```

jdalton:
Thanks, I'll dig a bit and report back.

jdalton:
Ok, I've pinged Brian Terlson for insight into the design choice.
This can be closed though since it would be yet another proposal (to remove order significance).

jdalton:
Just pinging back I made reify support unordered [export](https://github.com/benjamn/reify/blob/master/lib/parsers/acorn-extensions/export.js) and [import](https://github.com/benjamn/reify/blob/master/lib/parsers/acorn-extensions/import.js) specifiers.
It simplified the export extension a bit too ⚡️

loganfsmyth:
Sounds good. I don't think I'm comfortable landing any changes to this unless there were an official proposal, as you mentioned. I agree it's surprising behavior to me from a grammar definition standpoint, but being that this is the first time I've seen the question raised, I can't imagine it's a case that comes up often.

I think every time I've wanted a namespaced import and a named import, I've been fine to reference it off of the namedspace object instead. To me that says that this issue is more annoying for those writing tooling based around adding imports, since it makes adding imports more complicated, but it doesn't make it harder for day-to-day devs to get their work done.

jdalton:
>  I don't think I'm comfortable landing any changes to this unless there were an official proposal, as you mentioned. 

No worries 😎

>  I agree it's surprising behavior to me from a grammar definition standpoint, but being that this is the first time I've seen the question raised, I can't imagine it's a case that comes up often.
 
It's a bit WTFJS. I'm ok smoothing over the rough edge in my neck of the woods.

PR #532, opened 2017-05-22
Simplify PrivateName and ClassPrivateProperty types

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      |no
| Deprecations?     | no
| Spec compliancy?  | no
| Fixed tickets     | https://github.com/babel/babylon/pull/260#discussion_r117782754
| License           | MIT

Simplified the type declarations for `PrivateName` and `ClassPrivateProperty`.
* No need to mark `PrivateName` as `PatternBase`; it's not a pattern if it can't be used as a parameter.
* No need to define `__clone()` on it since that's never used.
* TypeScript doesn't support class private properties, and I'm pretty sure flow doesn't either, so no need to add ts/flow specific node declarations to `ClassPrivateProperty` just yet.

Just a thought: You might also consider using `ClassProperty` with a `PrivateName` key instead of creating a new node kind for that.

CC @diervo as #260 was just merged today.


andy-ms:
Actually, `ClasPrivateProperty` probably doesn't need `ClassMemberBase` either unless it can be static or have decorators.

hzoo:
Yeah Diego wanted it merged + transform before the TC39 meeting this week (Tuesday 😅 ) haha. Yeah maybe we should just remove any unnecessary types then given it's not supported anyway.

hzoo:
Although there is https://github.com/erights/Orthogonal-Classes but we can also just that stuff when we get to it (static, etc)

diervo:
Orthogonal classes have a lot of open questions and is not likely to be advancing any time soon.
Now regarding public and private fields, let's not worry about naming or final semantics until both proposals move forward.

Issue #538, opened 2017-05-26
Numeric Separator: Stage 1


hzoo:
> Advance Numeric separators proposal to Stage 1 slides (Sam Goto) (@samuelgoto)

## Info
Proposed at TC39 Meeting: [May 2017](https://github.com/tc39/agendas/blob/master/2017/05.md)
Slides at the meeting: https://docs.google.com/presentation/d/1hcajTemZB2Ruo4EePOyFiva1xpyv5ukKk4aQ0B83dUA/edit#slide=id.p
Proposal Repo: https://github.com/samuelgoto/proposal-numeric-separator

## Example
```js
var thousands = 10_000; // Instead of 10000.
var credit_card_number = 1234_5678_9012_3456; // Instead of 123456789012345.
var social_security_number = 999_99_9999; // Instead of 999999999.
var pi = 3.14_15; // Instead of 3.1415
var bytes = 0b11010010_01101001_10010100_10010010; // Instead of 0b11010010011010011001010010010010.
var 0xCAFE_F00D; // Instead of 0XCAFEF00D.
```

## Basic Rules
- to use the `_` character.
- only one consecutive underscore is allowed.
- only between digits (not allowed at the beginning or end of literals)

## ESTree/Parsing
- looks like we need to figure out how to change the tokenizer `tt.num`
- Start reading a number at `getTokenFromCode `: https://github.com/babel/babylon/blob/9709c8d932731273c7f936595491fe1568c48cc2/src/tokenizer/index.js#L456-L464
- calls `readNumber`, or `readRadixNumber` or `readInt`
- ultimately need the `value` to include the `_` or to just keep the regular value + track where the `_` are?

## Transform
- if the AST is simple (just the value/raw) would be trivial to remove (regex replace `_` with '')?

ESTree: I guess there's nothing to add to the AST spec since this changes the value of the number itself?

cc @ajklein and @domenic

rwaldron:
@hzoo I was thinking about writing a plugin for this, as a way to get involved. I will likely need your guidance :)

rwaldron:
I actually just saw this issue because I'm trying to setup the babylon repo locally, but it's failing when I run `npm install`, here's the failure: 


```
$ npm install

> babylon@7.0.0-beta.10 prepublish /Users/rwaldron/clonez/babylon
> cross-env BABEL_ENV=production yarn run build

yarn run v0.24.5
$ yarn run clean && rollup -c
yarn run v0.24.5
$ rimraf lib
[...]

PR #541, opened 2017-05-26
NumberLiteralSeparator: Stage 1 feature plugin. Closes gh-538

rwaldron:
Signed-off-by: Rick Waldron <waldron.rick@gmail.com>

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #538
| License           | MIT

I modified `readInt` to ignore `_` characters found in decimal, hex, binary and octal literals. It will also throw SyntaxError exceptions when the `_` is misused.


hzoo:
Amazing, that was super quick!

A few things to note:

- [ ] We need this to be under a flag like the other [plugins](https://github.com/babel/babylon#plugins). This just means creating a `hasPlugin('numericSeparators)` check so that this functionality isn't turned on by default.
  - You can check @diervo's PR as an example of this: https://github.com/babel/babylon/pull/260
- [ ] Update the readme to add that `numericSeparators` line to the plugins list
- [ ] Update the ast/spec.md (maybe no update needed)

rwaldron:
@hzoo thanks for the quick review turnaround, I will get these issues addressed and ping you when ready :)

hzoo:
Yeah, I should probably add that information to the contributing, let us know if there's anything that can be clarified or is frustrating

rwaldron:
> - [x] We need this to be under a flag like the other plugins. This just means creating a hasPlugin('numericSeparators) check so that this functionality isn't turned on by default.
>   - You can check @diervo's PR as an example of this: #260
> - [x] Update the readme to add that numericSeparators line to the plugins list
> - [] ~~Update the ast/spec.md (maybe no update needed)~~

@hzoo changes pushed. I didn't make changes to ast/spec.md because I don't believe there are any changes to make there (I could be wrong, but need guidance)

src/tokenizer/index.js
```diff
@@ -555,6 +572,23 @@ export default class Tokenizer extends LocationParser {
...
       let val;
+
+      if (this.hasPlugin("numericSeparator")) {
+        const prev = this.input.charCodeAt(this.state.pos - 1);
+        const next = this.input.charCodeAt(this.state.pos + 1);
+        if (code === 95) {
+          if (forbiddenNumericLiteralSeparatorSiblings.includes(prev) ||
+              forbiddenNumericLiteralSeparatorSiblings.includes(next) ||
+              Number.isNaN(next)) {
+            this.raise(this.state.pos, "Invalid NumericLiteralSeparator");
```
hzoo: I might actually create a different error message for multi `_`, if it's at the end or the beginning, or around invalid chars but we can do that later too.

src/tokenizer/index.js
```diff
@@ -555,6 +572,23 @@ export default class Tokenizer extends LocationParser {
     for (let i = 0, e = len == null ? Infinity : len; i < e; ++i) {
       const code = this.input.charCodeAt(this.state.pos);
       let val;
+
+      if (this.hasPlugin("numericSeparator")) {
+        const prev = this.input.charCodeAt(this.state.pos - 1);
+        const next = this.input.charCodeAt(this.state.pos + 1);
+        if (code === 95) {
+          if (forbiddenNumericLiteralSeparatorSiblings.includes(prev) ||
```
DrewML: Thanks for the PR, @rwaldron! I believe we need to switch to `Array#indexOf` here. We don't load any polyfills in Babylon, and the minimum supported node version is 4.0.0.

hzoo: Dang forgot about that 🤔

rwaldron: Ah, ok, no problem. I saw all the other modern JS and just assumed I could use it...whoops! Will fix now.

rwaldron:
@DrewML pushed change for includes -> indexOf

rwaldron:
> Ideally the cases that have a weird error message should be fixed to have the same error message as the others

I believe most those stay as they are because they have a different violation in them before the illegal `_`, or are already illegal and I wanted to ensure that the new code didn't erroneously consider them a new kind of restriction.  

`._` is illegal today for 


I think this:

> 0o01_8 is Unexpected token, expected ; (1:5) (module)

...is an outlier that should be fixed, but I'm not sure how. The `_` is legal here, the `8` is not—it should have an error that says `Expected number in radix 8`. Seems like an unrelated issue for another ticket. 

And this: 

> ._1_1 is Unexpected token (1:0) (module)

... involves changing `readToken_dot` to raise an exception, which I could do here or in a follow up PR. A follow up might be nicer to keep the error message fixes together.

hzoo:
Yeah I think this is good for this PR

hzoo:
Let's do this 😄

jridgewell:
Does `_123` pass?

diervo:
It shouldn't the separator is only allowed in between numeric values.

hzoo:
Well `_123` starts with _ so it should just be parsed as an Identifier right?

rwaldron:
@jridgewell @hzoo @diervo correct, `_123` is not valid _DecimalDigits_, because `_` is a valid _IdentifierStart_ character. I'm sure it's fine, but I will open a PR that adds a test which asserts that this remains true when the `numericSeparator` plugin is active.

michaelficarra:
@rwaldron I believe there's a bug in that this allows legacy octal integer literals to have separators. Even if this does restrict them, I don't see a test for that. Do you mind sending a PR to correct that?

michaelficarra:
Ping @rwaldron.

Issue #539, opened 2017-05-26
import.meta: Stage 2


hzoo:
> import.meta for stage 2 (Domenic Denicola) (@domenic)

## Info
Proposed at TC39 Meeting: [May 2017](https://github.com/tc39/agendas/blob/master/2017/05.md)
Spec Repo: https://github.com/tc39/proposal-import-meta
Spec Text: https://tc39.github.io/proposal-import-meta/

## Example

```js
(async () => {
  const response = await fetch(new URL("../hamsters.jpg", import.meta.url));
  const blob = await response.blob();

  const size = import.meta.scriptElement.dataset.size || 300;

  const image = new Image();
  image.src = URL.createObjectURL(blob);
  image.width = image.height = size;

  document.body.appendChild(image);
})();
```

## Parsing/ESTree
- [ ] make issue
- [x] Sounds pretty straightforward to make it a `MetaProperty` like `new.target`

```js
{
  "type": "MetaProperty",
  "meta": {
    "type": "Identifier",
    "name": "import"
  },
  "property": {
    "type": "Identifier",
    "name": "meta"
  }
}
```

- parsing `new.target` via `const metaProp = this.parseMetaProperty(node, meta, "target");` at 
 https://github.com/babel/babylon/blob/50694f99b12295b7515bb2c6c8f44b86a05deeca/src/parser/expression.js#L715
- dynamic import https://github.com/babel/babylon/blob/50694f99b12295b7515bb2c6c8f44b86a05deeca/src/parser/expression.js#L429

## Transform
- Not sure, maybe webpack should handle? Or in node?

Kovensky:
How should the tree look like when `MemberExpression` are involved, like in `import.meta.foo`?

```
// super-like
MemberExpression {
  object: MemberExpression {
    object: Import {}
    property: Identifier { name: 'meta' }
  }
  property: Identifier { name: 'foo' }
}
```
```
// new.target-like
MemberExpression {
  object: MetaProperty {
    meta: Identifier { name: 'import' }
    property: Identifier { name: 'meta' }
  }
  property: Identifier { name: 'foo' }
}
```

ESTree specifically defined an `Import` node expecting it to be used like `super` in the future.

https://github.com/estree/estree/blob/master/experimental/import-expression.md

bakkot:
I'd vote for making it like `new.target`, definitely. That's certainly how it's spec'd, and `import.somethingElse` is not grammatical.

[...]

PR #544, opened 2017-05-27
Add plugin for import.meta proposal

jkrems:
Fixes https://github.com/babel/babylon/issues/539

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #539
| License           | MIT

Relatively straight-forward: Parses `import.meta` as a new `MetaProperty`. The expression is valid in modules but not in scripts. It can be enabled even when `dynamicImport` isn't enabled.


jkrems:
The CI failure *seems* to be unrelated to this change..?

bakkot:
Should make sure to disallow `import.meta = val`. I don't think it's specced yet, but it will be.

jkrems:
Added tests to verify that:

1. `import.meta` is not a valid assignment target
2. `import.meta.prop` is a valid assignment target

test/fixtures/experimental/import-meta/without-dynamic-import/actual.js
```diff
@@ -0,0 +1,4 @@
+const x = import.meta;
+const url = import.meta.url;
+import.meta;
+import.meta.url;
```
hzoo: Was there a reason this test didn't also have ` import.meta.couldBeMutable = true;`?

jkrems: No deeper reason than "order of copy&paste". ^^ Copied the `actual.js`/`expected.json` over again to keep them in sync.

src/parser/expression.js
```diff
@@ -594,6 +598,16 @@ export default class ExpressionParser extends LValParser {
...
   }
 
+  parseImportMetaProperty(): N.MetaProperty {
+    const node = this.startNode();
+    const id = this.parseIdentifier(true);
+    this.expect(tt.dot);
+    if (this.options.sourceType !== "module") {
+      this.raise(id.start, "import.meta can only be used in modules");
+    }
+    return this.parseMetaProperty(node, id, "meta");
```
Kovensky: The error message if the meta property isn't `.meta` is confusing (it's hardcoded to mention `new`)

jkrems: Oops, good point. Added a test that checks for that & made it use the correct name.

src/parser/expression.js
```diff
@@ -588,12 +592,22 @@ export default class ExpressionParser extends LValParser {
...
     }
 
     return this.finishNode(node, "MetaProperty");
   }
 
+  parseImportMetaProperty(): N.MetaProperty {
+    const node = this.startNode();
+    const id = this.parseIdentifier(true);
+    this.expect(tt.dot);
+    if (this.options.sourceType !== "module") {
```
existentialism: Can also do `if (!this.inModule)` here

src/parser/expression.js
```diff
@@ -588,12 +592,22 @@ export default class ExpressionParser extends LValParser {
...
+      this.raise(id.start, "import.meta can only be used in modules");
```
existentialism: May want to make this consistent with other sourceType errors?

> 'import.meta' can only be used with 'sourceType: module'

src/parser/expression.js
jkrems: Weird, I thought I tried that first but `.inModule` was only true top-level. Must have missed something. You're right, definitely works fine.

src/parser/expression.js
jkrems: Heh, good point. Copied the wording from the "can only be used in functions". Should've taken the wording from import/export statements. Fixed.

rwaldron:
This might be irrelevant, or possibly a subject to have in a new issue, but I'm curious about the drops in coverage? For example, it says above that six files were impacted, but four of them aren't changed _at all_: 

- src/parser/comments.js	87.87% <0%> (-6.07%)	⬇️
- src/tokenizer/state.js	97.14% <0%> (-2.86%)	⬇️
- src/plugins/flow.js	97.35% <0%> (-1.04%)	⬇️
- src/tokenizer/index.js	97.41% <0%> (-1%)	⬇️

jkrems:
@rwaldron Yeah, I was wondering about the same thing. I tried rebasing on master in case it was using some weird merge head but afaict that wasn't it.

Issue #328, opened 2017-01-27
Optional Chaining: Stage 1


hzoo:
TC39 Proposals: https://github.com/tc39/proposals

Stage: 1, Gabriel Isenberg, Claude Pache

Proposed at TC39 Meeting: https://github.com/tc39/agendas/blob/master/2017/01.md#agenda-items
Spec Repo: https://github.com/tc39/proposal-optional-chaining
Slides: https://docs.google.com/presentation/d/11O_wIBBbZgE1bMVRJI8kGnmC6dWCBOwutbN9SWOK0fU/edit#slide=id.p

ESTree Issue to discuss AST Spec: https://github.com/estree/estree/issues/146

cc @gisenberg if you can implement or work with someone to do so. (Please join our slack if you haven't http://slack.babeljs.io/)

--- 
Babylon PR: https://github.com/babel/babylon/issues/545
Babel PR: https://github.com/babel/babel/pull/5786

gisenberg:
Thanks for looping me in! I wanted to let everyone know that I'm over in the BabelJS Slack if there are others that would like to discuss this feature further. I'm currently working with @bmeck on the Babel side of things.

kristofdegrave:
I'm currently werking on An PR for this

sunnylqm:
Any news? @kristofdegrave

rattrayalex:
As discussed in the babel slack, I'm also interested in contributing to this once the spec is a bit more settled. 

I previously implemented a `?.` feature for [LightScript](http://www.lightscript.org) (a fork/superset of JS built on babylon/babel). In case it helps anyone, here is: 

- The [babel transform](https://github.com/lightscript/babel-plugin-lightscript/blob/master/src/index.js#L932) which implements short-circuiting. 
- The [associated tests](https://github.com/lightscript/babel-plugin-lightscript/tree/master/test/fixtures/safe-member-expression)
- The [commit](https://github.com/lightscript/babylon-lightscript/commit/f59a365f1746e7a541fb29bcb82e7d770f186fb3) adding the feature to the lightscript parser (a fork of babylon). It uses poor naming ("elvis operator" isn't accurate) and isn't backwards-compatible; a different approach would be necessary. 

My version doesn't currently support assignments or delete, though I may add those soon.

cusspvz:
[...]

PR #545, opened 2017-05-28
Optional Chaining: Stage 1 plugin

xtuc:
> EDIT by hzoo: Check our [spec-new](https://github.com/babel/babylon/labels/spec-new) label to follow newly implemented proposals!

| Q                 | A
| ----------------- | ---
| Bug fix?          | n
| Breaking change?  | n
| New feature?      | y
| Deprecations?     | n
| Spec compliancy?  | y
| Tests added/pass? | y/n
| Fixed tickets     | #328
| License           | MIT

Based on the spec [here](https://github.com/claudepache/es-optional-chaining).

AST change:

```diff
interface MemberExpression <: Expression, Pattern {
  type: "MemberExpression";
  object: Expression | Super;
  property: Expression;
  computed: boolean;
+ optional: boolean | null;
}

interface CallExpression <: Expression {
  type: "CallExpression";
  callee: Expression | Super | Import;
  arguments: [ Expression | SpreadElement ];
+ optional: boolean | null;
}

interface NewExpression <: CallExpression {
  type: "NewExpression";
+ optional: boolean | null;
}
```

Edit:
### TODO
- [x] Add test for `a?.b.c.?.d`? (skipping ?. in between)


ast/spec.md
```diff
@@ -845,10 +845,11 @@ interface MemberExpression <: Expression, Pattern {
   object: Expression | Super;
   property: Expression;
   computed: boolean;
+  optional: boolean | null;
 }
 ```
 
-A member expression. If `computed` is `true`, the node corresponds to a computed (`a[b]`) member expression and `property` is an `Expression`. If `computed` is `false`, the node corresponds to a static (`a.b`) member expression and `property` is an `Identifier`.
+A member expression. If `computed` is `true`, the node corresponds to a computed (`a[b]`) member expression and `property` is an `Expression`. If `computed` is `false`, the node corresponds to a static (`a.b`) member expression and `property` is an `Identifier`. The `optional` flags indecates that the member expression can be called even if the object is null or undefined. If this is the object value (null/undefined) should be returned.
```
Kovensky: `indicates`

src/tokenizer/index.js
```diff
@@ -391,6 +391,18 @@ export default class Tokenizer {
     return this.finishOp(code === 61 ? tt.eq : tt.prefix, 1);
   }
 
+  readToken_question() { // '?'
+    const next = this.input.charCodeAt(this.state.pos + 1);
+    if (next === 46) { // 46 = question '.'
```
Kovensky: Comment is confusing; just `46 = '.'` should work.

src/parser/expression.js
```diff
@@ -301,6 +301,24 @@ export default class ExpressionParser extends LValParser {
         node.object = base;
         node.callee = this.parseNoCallExpr();
         return this.parseSubscripts(this.finishNode(node, "BindExpression"), startPos, startLoc, noCalls);
+
+      } else if (this.eat(tt.questionDot)) {
```
hzoo: Similar to the other PR, we need to add a flag. I added the steps in https://github.com/babel/babylon/blob/master/CONTRIBUTING.md#creating-a-new-plugin-spec-new.

`hasPlugin('nullPropagation')` or `hasPlugin('optionalChaining')`, etc. Probably `optionalChaining` since the repo was changed to that.

test/fixtures/experimental/optional-chaining/function-call/actual.js
```diff
@@ -0,0 +1 @@
+func?.()
```
hzoo: This test doesn't have an expected?

test/fixtures/experimental/optional-chaining/class-contructor-call/actual.js
```diff
@@ -0,0 +1 @@
+new C?.()
```
hzoo: This test doesn't have an expected?

src/tokenizer/index.js
hzoo: Or just `// '.'` like the other parts of the code

cpojer:
When I experimented with this about three years ago, I also went the route to make MemberExpressions optional. Seems like that's the way to go! :)

kerhong:
I believe this breaks valid syntax. The expression `foo?.1:bar` generates syntax error (Unexpected token), it should (and currently is) interpreted as `foo ? 0.1 : bar`.

src/parser/expression.js
xtuc: I added the plugin check, updated the corresponding documentation and also added an explicit error message since it's a new syntax.

test/fixtures/experimental/optional-chaining/function-call/actual.js
xtuc: Does parse yet, need to be implemented (i'm still WIP)

test/fixtures/experimental/optional-chaining/class-contructor-call/actual.js
xtuc: Does parse yet, need to be implemented (i'm still WIP)

hzoo:
@kerhong Yep, I mentioned this in my review and it's also in https://github.com/tc39/proposal-optional-chaining#notes.

src/parser/expression.js
```diff
@@ -301,6 +301,38 @@ export default class ExpressionParser extends LValParser {
...
+          node.computed = true;
+          this.expect(tt.bracketR);
+          base = this.finishNode(node, "MemberExpression");
+        } else if (this.eat(tt.parenL)) {
+          const possibleAsync = this.state.potentialArrowAt === base.start &&
+            base.type === "Identifier" &&
+            base.name === "async" &&
+            !this.canInsertSemicolon();
+
+          node.arguments = this.parseCallExpressionArguments(tt.parenR, possibleAsync);
```
jridgewell: There's no `callee`

jridgewell:
Opened a [PR](https://github.com/xtuc/babylon/pull/1) to complete the work, so hopefully this'll land soon.

Does anyone know why `.?` wasn't considered? As far as I can see, it fixes the `foo?.3:0` edge case without introducing any others.

xtuc:
Thanks for the PR @jridgewell, that's a nice job.

> Does anyone know why .? wasn't considered? 

Maybe because of the syntax? `a.?b` vs `a?.b`

jridgewell:
It's actually brought up several times:
- https://esdiscuss.org/topic/existential-operator-null-propagation-operator#content-11
- https://esdiscuss.org/topic/optional-chaining-aka-existential-operator-null-propagation#content-6

hzoo:
cc @yungsters @zertosh if you want to review

src/parser/expression.js
```diff
@@ -301,6 +301,44 @@ export default class ExpressionParser extends LValParser {
...
+          this.raise(startPos, "You can only use optional-chaining when the 'optionalChaining' plugin is enabled.");
+        }
+
+        if (noCalls && this.lookahead().type == tt.parenL) {
+          return base;
+        }
+        this.next();
+
+        const node = this.startNodeAt(startPos, startLoc);
+
```
hzoo: Could probably just move `node.optional = true;` to here since it's in all the if conditionals?

jridgewell: I was trying to keep the property ordering consistent between optional and non-optional branches. See the `#parseNew` as well.

src/tokenizer/index.js
```diff
@@ -426,6 +426,18 @@ export default class Tokenizer extends LocationParser {
     return this.finishOp(code === 61 ? tt.eq : tt.prefix, 1);
   }
 
+  readToken_question() { // '?'
+    const next = this.input.charCodeAt(this.state.pos + 1);
+    const next2 = this.input.charCodeAt(this.state.pos + 2);
+    if (next === 46 && !(next2 >= 48 && next2 <= 57)) { // '.' not followed by a number
```
ljharb: what are these magic numbers? could these be stored in named variables, so it's clear what "46" means?

xtuc: These are char codes. I'll add some contants to improve readability, thanks.

davidyaha:
🎉 🎉

hzoo:
Nice work everyone! Really appreciating the reviews!

Babel transform PR is https://github.com/babel/babel/pull/5786 (please check that out next)

jridgewell:
Actually, the transform PR is https://github.com/babel/babel/pull/5813.

Issue #536, opened 2017-05-24
Arrow function followed by slash


Ginden:
Related to [this discussion](https://esdiscuss.org/topic/arrow-function-followed-by-divide-or-syntax-error).

**Input**:
```
x=x=>{}/alert(1)/+alert(2)// alerts on edge, syntax error on other browsers
```

**Result**: parses succesfully.

**Expected result**: SyntaxError

hzoo:
Hey @Ginden! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

nicolo-ribaudo:
Also `() => {} ? a : b`.

Can I fix this issue?

existentialism:
@nicolo-ribaudo go for it!

PR #559, opened 2017-06-03
Arrow fns can't be used as the left side of a binary or ternary expression

nicolo-ribaudo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #536 
| License           | MIT

I followed the same way used to forbid `() => {}()`: when parsing a binary/ternary expression, if an arrow function is encountered it is immediatly returned.

I didn't add the same check to `parseMaybeUnary` because this error is already caught when checking if the expression can be incremented/decremented. Should I add there the same check?

---

This is my first contribution to babel/babylon: do I have to sign a contributor license agreement?


danez:
Will these examples still work?
```js
(() => {}) || true ;
(() => {}) ? a:b;
```

acorn fixed this with https://github.com/ternjs/acorn/commit/6a2de6dcfd4873a98614f36c5cc259b04af03b0a

nicolo-ribaudo:
They will work because `potentialArrowAt` will be the start of the first parenthesis, so it will be different from the start of the arrow function.
Anyway, I won't have time to add the tests until the 13th.

danez:
Ah nice. I add the two tests and merge it then. Thank you.

nicolo-ribaudo:
Thank you for taking care of adding the two tests.

PR #565, opened 2017-06-04
Fix location info on FunctionTypeParam nodes

xixixao:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     |
| License           | MIT

The location info was wrong since these got parsing support in Novemeber (in #197 ). This is in turn breaking Recast, which is in turn breaking `jscodeshift` on programs using Flow function param annotations without names.

I am a complete n00b here, how long will it take for this fix to make it to npm and what version will it go out in? (can it be patched on 6?). Thanks!

### DIFF
--- a/src/plugins/flow.js
+++ b/src/plugins/flow.js
@@ -708,7 +708,7 @@ export default (superClass: Class<Parser>): Class<Parser> => class extends super
   }
 
   reinterpretTypeAsFunctionTypeParam(type: N.FlowType): N.FlowFunctionTypeParam {
-    const node = this.startNodeAt(type.start, type.loc);
+    const node = this.startNodeAt(type.start, type.loc.start);
     node.name = null;
     node.optional = false;
     node.typeAnnotation = type;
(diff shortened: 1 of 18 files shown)


danez:
Thank you very much.
This is funny that this could happen, because flow should have catched this error easily. Seems our flow types are a little broken. I will look into this and report back.

As soon as this is merged I can make a patch release for v6

hzoo:
Lol yeah we should figure out how to fix that 😛

hzoo:
@andy-ms any ideas?

danez:
The issue is this line. https://github.com/babel/babylon/blob/master/src/types.js#L38 It makes every node to be this type:
```js
type Node = { [string]: any };
```

Seems that defined fields and typeindexers cannot be used together in flow. 
But if i remove this line I get around 300 errors. which in most cases is just missing type casting and flow.

xixixao:
@danez Can I help backporting this to 6x or is it good to go?

danez:
I released 6.17.3 with the change.

PR #566, opened 2017-06-05
Fix parsing of private fields

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

Do not parse computed and literal keys for PrivateClassProperties. By looking at the spec it should always be an Identifier afaics. This also removes the computed key as it will always be false, and it is not mentioned in the spec.

so this should not work I think:

```js
class B {
  #"x";
  #1;
  #["as"];
}
```

Little unrelated:
I also noticed that our spec and types say, that key for ClassProperties (non-private) is an Identifier, but the current official spec says it is: `StringLiteral | NumericLiteral | Identifier | ComputerProperty` so I went with `Expression`, same as ClassMethod.

This all parses fine and I think it is correct:
```js
class A {
  "b" = 1;
  3=4;
  ["as"+"s"];
}
```


test/fixtures/experimental/class-private-properties/pbn-success.1/actual.js
```diff
@@ -1,19 +0,0 @@
...
-    get x() { return this.#x }
-    set x(value) { this.#x = +value }
-
-    get y() { return this.#y }
-    set y(value) { this.#y = +value }
-
-    equals(p) { return this.#x === p.#x && this.#y === p.#y }
-
-    toString() { return `Point<${ this.#x },${ this.#y }>` }
-}
```
danez: This test was duplicate

test/fixtures/experimental/class-private-properties/asi-failure-computed/actual.js
```diff
@@ -1,4 +1,4 @@
 class Foo {
   #p = x
-  [#m] () {}
```
danez: This would have been fine actually, besides the method. This looks totally valid to me:
```js
class A {
  #x = 1 // private field
  [#x] = 2 // public field
}
```

diervo: Have you created a test for this?

diervo:
@littledan should review the expression, we might be allowing more use cases than the expect intended

test/fixtures/experimental/class-private-properties/asi-failure-literal/actual.js
```diff
@@ -0,0 +1,4 @@
+class Foo {
+  #"p" = x
+  #2 = y
```
littledan: It'd be nice to break this out into a separate test file, so you can check that each of them is a syntax error.

littledan:
@diervo Looks like the key for a ClassMethod is an Expression. This should be the same as whatever the key of a class method is. With hindsight, Babel might've gone for a more restrictive choice that more closely encodes the grammar, so that transforms can operate with a little more confidence and determinism, but Expression is a reasonable choice that expresses the possibilities.

test/fixtures/experimental/class-private-properties/asi-failure-computed/actual.js
danez: I think my example should throw, by looking at it again.

What I meant is this:
```js
class A {
  #x = 1 
  [this.#x] = 2
  //or
  #4 = this.#x
}
```

But these are also early errors?

littledan: `#4` should be a syntax error, as `#` needs to be followed by an IdentifierName.\

`[this.#x]` should be a ReferenceError when the class declaration is executed, as `#x` is "in scope" as a name during the evaluation of computed property names, but then `#x` cannot possibly be present on `this`.

danez: Upps, the #4 was not intended to be there. Basically I wanted to show initializing a private field from another private filed.

PR #573, opened 2017-06-10
Distinguish between ternary's : and arrow fn's return type

nicolo-ribaudo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | #58 
| License           | MIT

I took this approach:
1. After `?`, try to parse an assignment expression (consequent)
2. If the function threw:
  a. Disallow arrow functions in the first position
  a. Parse the assignment expression again (consequent)
3. Else, it there isn't `:`
  a. Find every typed arrow function in that expression
  b. If there are more than one, trow an "Ambiguity" error ¹
  c. Mark the arrow function so that it is parsed as `ParenthesizedExpression : ArrowFunctionExpression`
  d. Parse the assignment expression again (consequent)
4. Parse `:`
5. Parse an assignment expression (alternate)

---

- [x] ~~~[1] I don't know if the "check if there are more than one functions" condition is ok, I have to think about it~~~
- [x] This approach doesn't work sometimes, like with `a ? (b) : c => (d => e) : f => g;`. I think I have to rewrite 3.b and 3.c


nicolo-ribaudo:
> - [x] I don't know if the "check if there are more than one functions" condition is ok, I have to think about it

I'm not 100% sure, but I couldn't find any example for which this check is wrong

test/fixtures/flow/regression/issue-58-ambiguous/actual.js
```diff
@@ -0,0 +1,4 @@
+// This can be parsed in two ways:
+// a ? b : (c => ((d): e => f))
+// a ? ((b): c => d) : (e => f)
+a ? (b) : c => (d) : e => f;
```
nicolo-ribaudo: Here throwing is the most future-compatible behavior. Probably we should ask to the flow team how to handle this cases.

vjeux:
Would be awesome if you could add https://github.com/prettier/prettier/issues/2194#issuecomment-309499002 as a test case. Thanks so much for fixing a bug before it was even discovered in prettier!

nicolo-ribaudo:
@vjeux Test added 👍
 
---

@&#8203;anyoneFromTheBabelTeam Do you prefer that I resolve the merge conflicts by rebasing or by merging `master` into this branch?

vjeux:
@nicolo-ribaudo awesome, thanks a ton! I can't wait for it to ship!

hzoo:
@nicolo-ribaudo rebase although if it's a simple change with 1 author we just squash merge it anyway

vjeux:
Fyi, this bug was also found by a Facebook employee during a conversion. So it's at least impacting two people :)

nicolo-ribaudo:
@existentialism I'm trying to remove that parameter, but this refactor is not ready yet and I can't guarantee I will succeed: this PR could be merged with that ugly parameter and I'll open another one when it will be ready

nicolo-ribaudo:
@existentialism I managed to remove that parameter!

existentialism:
@nicolo-ribaudo nice! @hzoo and I were discussing using state for this vs a param a while back too!

Can also think about using a set for `state.noArrowParamsConversionAt` at some point (to avoid lots of potential indexOf checks).

But for now, LGTM 👍

nicolo-ribaudo:
> Can also use a set for state.noArrowParamsConversionAt (to avoid lots of potential indexOf checks)?

I searched in the babylon source code and it doesn't seem to use any ES6 function: wouldn't using `Set` reduce babylon's compatibility?
Another option to replace `indexOf` might be to use a function which traverses `state.noArrowParamsConversionAt` from the end and returns early if there is a number which is less than the searched one (sice `state.noArrowParamsConversionAt` is sorted)

Anyway, I think `indexOf` is fine because the array will never contain many items (unless there are lots of nested `?:` and `() =>`), so its performance cost should be negligible.

hzoo:
The initial purpose of Babel 7 was just to drop Node 0.10/0.12. We are using Set in other parts of Babel already and it helped us to drop babel-runtime as dep (Symbol as well), so that's not an idea. But yeah if those specific cases won't have a lot of items then we don't have to change it.

hzoo:
Actually was getting a lot of errors locally for flow `npm run flow` so going to revert for now and release with what we have. Can you make another PR to fix? I'm not sure why the flow tests are optional (maybe since the infra setup is still broken) https://travis-ci.org/babel/babylon/jobs/246895914

nicolo-ribaudo:
Errors fixed in #596

nicolo-ribaudo:
> I'm not sure why the flow tests are optional

Only the flow test suite is optional, not the type checks. Actually it looks like they pass on travis-ci: 

https://travis-ci.org/babel/babylon/jobs/246895911
> The command "if [ "$JOB" = "test" ]; then yarn test-only; fi" exited with 0.
11.33s$ if [ "$JOB" = "lint" ]; then yarn run lint && yarn run flow; fi
yarn run v0.24.6
$ eslint src bin 
yarn run v0.24.6
$ flow 
Launching Flow server for /home/travis/build/babel/babylon
Spawned flow server (pid=2240)
Logs will go to /tmp/flow/zShomezStraviszSbuildzSbabelzSbabylon.log
No errors!

hzoo:
Oh ok that was confusing then

Issue #574, opened 2017-06-11
6.17.2 release breaks /* istanbul ignore */ comments in some scenarios


bcoe:
[this change](https://github.com/babel/babylon/pull/478) introduced in babylon@6.17.2 seems to break Istanbul's ignore next functionality in some situations. See the following code:

### Input Code

```js
'use strict'

const test = foo => {
  switch (foo) {
    case 'ok':
      return console.log('this is fine')

    /* istanbul ignore next */
    default:
      throw new Error('nope')
  }
}

test('ok')
```

### Expected Behavior

the comment ` /* istanbul ignore next */` ignores the `default` statement (behavior in 6.17.1).

### Current Behavior

the comment `/* isanbul ignore next */` fails to ignore the default switch statement (behavior in 6.17.2).

### Thoughts

I believe the issue is that the comment gets folded up onto the prior line, and then we fail to traverse to the `default` statement when deciding what to ignore.

CC: @aardito2

hzoo:
Hey @bcoe! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

loganfsmyth:
Have you considered using https://github.com/istanbuljs/babel-plugin-istanbul? Generally I don't think I'd recommend running tooling on the output from Babel.

bcoe:
@loganfsmyth https://github.com/istanbuljs/babel-plugin-istanbul/blob/master/package.json#L13 this change landed in `babylon@6.17.2 ` breaks `babel-plugin-istanbul` and `istanbuljs`.

bcoe:
It seems like the underlying issue might be that the comment will no longer be treated as a `leadingComment` as a result to moving changing its scope, which breaks this logic in istanbul-lib-instrument:

```js
            node.leadingComments.forEach(function (c) {
                const v = (c.value || /* istanbul ignore next: paranoid check */ "").trim();
                const groups = v.match(COMMENT_RE);
                if (groups) {
                    hint = groups[1];
                }
            });
```

loganfsmyth:
Ahh sorry, I misunderstood the issue then.

aardito2:
[...]

PR #575, opened 2017-06-12
Fix comment attachment for call expressions

aardito2:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | https://github.com/babel/babylon/issues/574
| License           | MIT

Fix for incorrect comment attachment for `CallExpression` introduced in https://github.com/babel/babylon/pull/478

If a `CallExpression` is not followed by a semicolon and has a comment following it, the comment will now be attached as a trailing comment on the parent node rather than incorrectly being added to the trailing comments of the `CallExpression`'s arguments.


bcoe:
Thanks for digging into this so fast; I'm definitely pretty out of my element with the parser bits of Babylon, hope to learn more.

danez:
Thanks, not sure if having all the special cases for nodes in there is the way to go, but this only fixes the current code 👍

aardito2:
I definitely agree - including special cases for certain node types works for now, but there should be a better, more general solution for handling these cases going forward.

Issue #456, opened 2017-04-07
Babylon doesn't parse for of regex


vjeux:
```js
for (a of /a/) {}
```

```js
SyntaxError: Unexpected token (1:10)
> 1 | for (a of /a/) {}
    |           ^
```

But it seems to parse both on flow and Chrome:

![image](https://cloud.githubusercontent.com/assets/197597/24816731/ab34fb0a-1b8e-11e7-9bd4-124902dccf45.png)

hzoo:
Hey @vjeux! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

PR #579, opened 2017-06-15
Alow regex after "of" in for(...of...)

nicolo-ribaudo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #456 
| License           | MIT

I initially tried to set `this.state.exprAllowed` in the parser, after eating the "of" word ([here](https://github.com/babel/babylon/blob/50ae16de38150caf84988b80098df93173f718b9/src/parser/statement.js#L544)), but then I decided to fix this bug directly in context.js.

This fix, teorically, has a "bug": in `for (a++ of /foo/g);`, the regex is parsed as a division because `++` can be followed by an expression. In practice this can't happen, because `a++` is not a valid assignment target and thus babel throws before encountering `/`.


existentialism:
This looks fine, but we may also want to ref [acorn's fix](https://github.com/ternjs/acorn/commit/ab112310949717879c9e67bbf7d907c3dcac38b5), as there's also another case of handling contextual yield.

nicolo-ribaudo:
The bug with yield fixed by that commit already works in Babylon: https://astexplorer.net/#/gist/806e25352b11e858e13a0ccaf722816f/latest

PR #581, opened 2017-06-16
Fix v8 deopts

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | n
| Breaking change?  | n
| New feature?      | n
| Deprecations?     | n
| Spec compliancy?  | n
| Tests added/pass? | y
| Fixed tickets     | 
| License           | MIT

This fixes some v8 deopts, the biggest one is the loop change in `skipLineComment()`. It was deoptimized because` this.input.charCodeAt(this.state.pos);` was triggering an `out of bounce deopt`.

The removed `clone()` seems to be unnecessary to me?

The rest is about hidden class deopts, although I still see deopts for `this.state` in the beginning of the run, but that is normal? At least I couldn't find more dynamic fields which are not initialized.

The overall performance did not change with this changes.

### DIFF
--- a/src/tokenizer/index.js
+++ b/src/tokenizer/index.js
@@ -78,6 +78,7 @@ export default class Tokenizer extends LocationParser {
     super();
     this.state = new State;
     this.state.init(options, input);
+    this.isLookahead = false;
   }
 
   // Move to the next token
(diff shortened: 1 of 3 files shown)


src/tokenizer/index.js
```diff
@@ -127,7 +128,7 @@ export default class Tokenizer extends LocationParser {
     this.next();
     this.isLookahead = false;
 
-    const curr = this.state.clone(true);
```
hzoo: just not sure about this one

danez: Yes, but isn't one clone enough?

hzoo: oh I didn't see the ^ lol

PR #583, opened 2017-06-17
update babel alpha 12, fixes build issues

hzoo:


danez:
Maybe we should also turn of the codecov PR comments:

- It posts multiple times recently
- The information is not useful and als available in the PR checks

hzoo:
Yeah I basically ignore them now..

PR #587, opened 2017-06-17
Decorators Stage 2 Parsing

peey:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #13, #33
| License           | MIT
 

1. Adds a `decoratorsStage2` plugin which can be used instead of `decorators` plugin but both cannot be used together
2. Does not interop with class fields. This functionality will be added later when the [class fields](https://github.com/tc39/proposal-class-fields) and [unified class fields](https://github.com/littledan/proposal-unified-class-features) advance 
3. Disallows the following
  3.1. Computed key on decorator e.g. `@dec[foo]`
  3.2. Parameter decorators
  3.3. Object methods
  3.4. `export` keyword in between of class body and decorators
4. Allows the following
  4.1 Decoration of computed class method names (#33)
  4.2 Decoration of generator functions (#13)

In some places, I've chosen to throw custom errors like "Stage 2 decorators do not allow parameter decoration" instead of throwing unexpected token so as to help people migrate and to make these changes explicit and less confusing.

## Rationale for PR
The disallowed features are much loved but unfortunately either aren't a part of the proposal yet or have been discarded, therefore there is no clear answer on how to transform them. 

This PR and associated plugin for stage-2 decorators is going to be opt-in for those who want to start writing what's currently stage-2 approved in the decorators proposal.

Ping @bakkot - #353; @DrewML - #236


nicolo-ribaudo:
I don't like the `decoratorsStage2` name for the plugin: what will happen when the decorators proposal reaches stage 3? I think names like `decoratorsNew` or `decorators2` may be good candidates.

peey:
That's a valid point. We could also name it `classAndFieldDecorators` like it was named in #236 if we don't mind tying it to that feature (then we'll have to put future decorator proposals like [parameter decorators](https://docs.google.com/document/d/1Qpkqf_8NzAwfD8LdnqPjXAQ2wwh8BBUGynhn-ZlCWT0/edit) under a different plugin name)

hzoo:
@nicolo-ribaudo depends on if stage 3 means more changes. Either way we would need to fork the logic again with a new plugin/name

hzoo:
@vjeux brought up a good point about whether the start position of the class/function part of the decorator should change? https://twitter.com/Vjeux/status/876213458622005249

peey:
@hzoo Though in case stage 3 brings mostly non-breaking changes, we could easily make additions to this new decorators plugin instead of forking logic into a new one.

So we should definitely choose a plugin name that's not tied to stage 2

basicdays:
Would a test to assert the output of the following make sense?

```js
export default @thing
class Stuff { }
```

peey:
```
export default @thing
class Stuff { }
```
@basicdays yep, I think this is the preferred way to do it since [decorators and class tail together make up the class](https://tc39.github.io/proposal-decorators/#prod-ClassDeclaration), and then you can export that like you export an undecorated class

I will add a test for this

src/index.js
```diff
@@ -62,6 +62,10 @@ function getParserClass(pluginsFromOptions: $ReadOnlyArray<string>): Class<Parse
     pluginList.unshift("estree");
   }
 
+  if (pluginList.indexOf("decorators") >= 0 && pluginList.indexOf("decoratorsStage2") >= 0) {
```
hzoo: Don't have to do this now, but we should figure out a better solution to this in the future (regarding multiple versions of the same kind plugin). Same kind of error as TS + Flow

test/fixtures/experimental/decorators-stage-2/no-computed-decorator-member/actual.js
```diff
@@ -0,0 +1,6 @@
+class Foo {
+  @bar[bizz]
```
littledan: This is a great test; you could also add tests for cases like `@(bar.baz)` and `@foo().bar`, which one might expect to work. The grammar is extremely restrictive here.

peey:
Changes in latest commit
1. Add test for exports as @basicdays suggested
2. Add one test case for class expressions, e.g. `var foo = class {}`
3. Add tests to ensure `@(bar.baz)` and `@foo().bar` fail as @littledan suggested
4. Minor fix (use `pluginsFromOptions` instead of `pluginList`) for code that detects plugin conflict, and added a test

I think I'll do the plugin renaming in the end after everything else in the PR is done and approved

hzoo:
> What's your long-term vision for this mode? Ultimately, it would be great if, in Babel 7.0 (or 8.0), this new syntax (and a new matching transform) would be the only decorators implementation, or at least the only one selected by the staged presets, but maybe this is too big of a breaking change.

Yeah we would want the preset to always be updated to the latest version (via major bumps in both the babel transform and the babel stage preset) and babylon itself would update with minor bumps via new plugin names

peey:
> @vjeux brought up a good point about whether the start position of the class/function part of the decorator should change? https://twitter.com/Vjeux/status/876213458622005249

@vjeux trying to find the reason for this in the proposals document. Is it because the productions look like

```
ClassDeclaration[Yield, Default]:
    DecoratorList[?Yield]opt class BindingIdentifier[?Yield] ClassTail[?Yield]
    [+Default]DecoratorList[?Yield]opt class ClassTail[?Yield]
```

So we'd consider ClassDeclaration to start at the begining of the first decorator? (And similarly for [class expression](https://tc39.github.io/proposal-decorators/#prod-ClassExpression) and [methods](https://tc39.github.io/proposal-decorators/#prod-ClassElement)) ??

vjeux:
All the other parent nodes include the location of the children node except for decorators. We use this property in prettier to attach comments by going down the ast looking at ranges, but it breaks for decorators. If you could fix it that would remove one hack we have to do.

hzoo:
Yeah the reason is just that you'd have to backtrack to know the start of the "decorated node"

peey:
@hzoo I've found an easy way to do it though using `this.resetStartLocationFromNode` method, so it won't be a problem to add this feature. Just debugging flow errors right now before I update the PR!

hzoo:
Cool didn't know about that one 😄

peey:
Latest changes in last 3 commits: 
 - Now all decorated elements (classes and methods) have their start location at the start location of the first decorator
 - Decide to rename to the plugin to `decorators2` instead of other variants because
   1. If more backwards-incompatible changes are introduced in future and then a new plugin may be implemented as `decorators3`
   2. Can rename it to `standardDecorators` at zero cost later when the proposal is at stage 4

These two edits have caused a lot of minor changes to a lot of files so it might be a pain to review. The only logic changes to source folder is by [this commit](https://github.com/babel/babylon/pull/587/commits/e1972baff51e838f226b6063026ed83e992b411a) in the file  `src/parser/statement.js` and everything else is renaming / test changes and I don't think would need review.

I also can't figure out how to increase coverage, the uncovered lines aren't making any sense to me (they existed in codebase before this PR). So if that's not a problem then I think we're good to merge (perhaps after latest changes relating to start location have been reviewed?)

src/index.js
```diff
@@ -47,6 +47,11 @@ const parserClassCache: { [key: string]: Class<Parser> } = {};
 
 /** Get a Parser class with plugins applied. */
 function getParserClass(pluginsFromOptions: $ReadOnlyArray<string>): Class<Parser> {
+
+  if (pluginsFromOptions.indexOf("decorators") >= 0 && pluginsFromOptions.indexOf("decorators2") >= 0) {
```
Kovensky: I'd say to use `Array.prototype.includes` instead...... but that doesn't work on Node 4 :cry:

test/fixtures/experimental/decorators-2/no-computed-decorator-member/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "plugins": ["decorators2", "classProperties"],
+  "throws": "Stage 2 decorators may only be used with a class or a class method (2:2)"
```
hzoo: Should this error message mention no computed instead?

peey: We will have to explain this in documentation or some other way. 

the problem is that with classProperties enabled and due to the fact that the computed property isn't considered a part of the decorator, we can't really tell if the person is trying to access decorator's member or trying to decorate a computed class property. 

So in this example it thinks that `[bizz]` is a class property.

If you disable class properties altogether you will get a syntax error at abc since it is expecting opening bracket right after the method name `[bizz]`

test/fixtures/experimental/decorators-2/no-export-decorators-on-class/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "sourceType": "module",
+  "throws": "Leading decorators must be attached to a class declaration (2:0)"
```
peey: good idea

hzoo:
Followup PR https://github.com/babel/babylon/pull/590

Issue #569, opened 2017-06-06
BigInt: Stage 2


littledan:
Champion: @littledan 
Spec Repo: https://github.com/tc39/proposal-bigint
Spec Text: https://tc39.github.io/proposal-bigint
Slides (to Stage 3 at May TC39 Meeting): https://docs.google.com/presentation/d/1lrcjQzIFgdUXczeeAzs4GkTXJsRQU21UhtmXef70Qic/edit#slide=id.p

## Parser

- Similar to https://github.com/babel/babylon/issues/538, we'll need to modify the tokenizer to accept `10n` 
as a valid number. (Check that [PR](https://github.com/babel/babylon/pull/541))
- From a syntactic perspective, the only thing added is numeric literals of the form `987235987032n`. These permit hexadecimal, octal and binary prefixes, but no exponential form, and no decimal place. See the grammar [in this section of the proposed spec](https://tc39.github.io/proposal-bigint/#sec-grammar-change).

## Transform

- All arithmetic operations and comparisons should be replaced by calls to functions which provide the appropriate behavior depending whether arguments are BigInts or Numbers.
  - The transform is slow, so it will really be more of a toy rather than something that should be used in production. For this reason, I'd argue it should be left out of Babel presets.
- Polyfill needed--ideally, this would be a thin wrapper around another BigNum library from npm. There are just a few methods to support. Some functions will have to be a bit different to get close to the right behavior, for example, `valueOf()` should throw a TypeError, so that implicitly casting to Number throws a TypeError, even if it would really return the BigInt. Additionally, the polyfill needs BigInt64Array and DataView operations.

hzoo:
Hey @littledan! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

wdhorton:
I'd like to take on this issue!

amilajack:
> All arithmetic operations and comparisons should be replaced by calls to functions which provide the appropriate behavior depending whether arguments are BigInts or Numbers.

[...]

PR #588, opened 2017-06-18
Stage 2: BigInt

wdhorton:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | #569
| License           | MIT

For the tokenizer, check if the last character is `n`, in which case it should be treated as a BigInt. If it has a dot or an E, throw, otherwise, advance the position one character. Replace the literal `n` with "" when parsing the value, and create a new token type `bigint`.

For the parser, when handling a `bigint` token, create a node of the new type `BigIntLiteral` with the given value.


wdhorton:
@littledan would love for you to take a look, let me know what other test cases I should add

hzoo:
Awesome @wdhorton!!

As for test cases can you add hex, octal, binary and then error for exponential form (you already have the error for decimals)?

test/fixtures/experimental/bigint/invalid-0/actual.js
```diff
@@ -0,0 +1 @@
+1.0n
```
hzoo: can rename the invalid-0 to just invalid-decimal?

README.md
```diff
@@ -142,6 +142,7 @@ require("babylon").parse("code", {
  - `numericSeparator` ([proposal](https://github.com/samuelgoto/proposal-numeric-separator))
  - `optionalChaining` ([proposal](https://github.com/tc39/proposal-optional-chaining))
  - `importMeta` ([proposal](https://github.com/tc39/proposal-import-meta))
+ - `bigint` ([proposal](https://github.com/tc39/proposal-bigint))
```
hzoo: What do we think about making the plugin name `bigInt` with camelCase?

src/tokenizer/index.js
```diff
@@ -663,9 +664,17 @@ export default class Tokenizer extends LocationParser {
...
+        if (isFloat) this.raise(start, "Invalid BigIntLiteral");
+        ++this.state.pos;
+        isBigInt = true;
+      }
+    }
+
     if (isIdentifierStart(this.fullCharCodeAtPos())) this.raise(this.state.pos, "Identifier directly after number");
 
-    const str = this.input.slice(start, this.state.pos).replace(/_/g, "");
+    const str = this.input.slice(start, this.state.pos).replace(/[_n]/g, "");
```
hzoo: Is it worth putting a comment here to explain we are removing the _ for numeric separators and n for BigInt?

README.md
wdhorton: I actually started with the plugin named bigInt. But then when it came time to make a new token type, I saw that the `RegExp` constructor mapped to `regex` so I figured that `BigInt` should map to `bigint`. So then I changed the plugin to match the token type.

But I'm ok with `bigInt` if you think that's better

src/tokenizer/index.js
wdhorton: sound like a good idea

README.md
hzoo: Yeah I just meant for the plugin string not the `tt.bigint` in the code if that makes sense (I wasn't clear earlier 😄)

src/tokenizer/index.js
```diff
@@ -678,7 +687,7 @@ export default class Tokenizer extends LocationParser {
     } else {
       val = parseInt(str, 8);
```
littledan: None of these parseInt calls are appropriate if isBigInt, as they will round away part of the answer.

README.md
hzoo: Yep that's what I meant!

test/fixtures/experimental/bigint/valid-0/expected.json
```diff
@@ -0,0 +1,69 @@
...
+            "end": {
+              "line": 1,
+              "column": 4
+            }
+          },
+          "extra": {
+            "rawValue": 100,
+            "raw": "100n"
+          },
+          "value": 100
```
hzoo: rawValue is basically the same as value and raw is just the actual string of the code

```
    this.addExtra(node, "rawValue", value);
    this.addExtra(node, "raw", this.input.slice(startPos, this.state.end));
```

wdhorton: @littledan I realized after I pushed this up that using a number as the value would defeat the purpose, since as you said we still couldn't represent 64-bit ints without rounding. I'll change it to strings

src/tokenizer/index.js
wdhorton: you're very right, that was an oversight on my part

src/tokenizer/index.js
```diff
@@ -631,10 +631,27 @@ export default class Tokenizer extends LocationParser {
...
   readRadixNumber(radix: number): void {
+    const start = this.state.pos;
+    let isBigInt = false;
+
     this.state.pos += 2; // 0x
     const val = this.readInt(radix);
     if (val == null) this.raise(this.state.start + 2, "Expected number in radix " + radix);
+
+    if (this.hasPlugin("bigInt")) {
+      if (this.input.charCodeAt(this.state.pos) === 110) { // 'n'
```
mathiasbynens: How about using `0x6E` instead of `110`? It makes it more obvious that this refers to U+006E LATIN SMALL LETTER N.

src/tokenizer/index.js
```diff
@@ -644,6 +661,7 @@ export default class Tokenizer extends LocationParser {
     const start = this.state.pos;
     let octal = this.input.charCodeAt(start) === 48; // '0'
```
mathiasbynens: Same could be done here (`0x30` instead of `48`)

src/tokenizer/index.js
```diff
@@ -661,11 +679,27 @@ export default class Tokenizer extends LocationParser {
       if (next === 43 || next === 45) ++this.state.pos; // '+-'
```
mathiasbynens: `if (next === 0x2B || next === 0x2D)`

src/tokenizer/index.js
```diff
@@ -661,11 +679,27 @@ export default class Tokenizer extends LocationParser {
...
       if (this.readInt(10) === null) this.raise(start, "Invalid number");
       isFloat = true;
+      next = this.input.charCodeAt(this.state.pos);
+    }
+
+    if (this.hasPlugin("bigInt")) {
+      if (next === 110) { // 'n'
```
mathiasbynens: Same here

test/fixtures/experimental/bigint/valid-binary/expected.json
```diff
@@ -0,0 +1,69 @@
...
+            "end": {
+              "line": 1,
+              "column": 12
+            }
+          },
+          "extra": {
+            "rawValue": "0b101011101",
+            "raw": "0b101011101n"
+          },
+          "value": "0b101011101"
```
littledan: Should Babylon handle converting this binary value into a decimal value, or should the transform be responsible for that/

src/tokenizer/index.js
wdhorton: The existing code was already checking the charCodes using non-hex integers, so I was just trying to stay consistent with that. I can go ahead and change it if you think that's best

test/fixtures/experimental/bigint/valid-binary/expected.json
wdhorton: I was thinking about that, and on a practical level I think it makes sense to delay that until the transform. If there was a binary value that exceeded the max value for a number, then in order to convert it we'd have to import a bigint library here. Whereas with the transform we're already going to need to use a library for the operations, so it makes sense to have it do the conversions as well

hzoo: Would this be the same question as for Numeric Separators? We have a `babel-plugin-transform-es2015-literals` that does binary conversion but only for regular numbers. 

https://github.com/babel/babel/pull/5793/files#diff-16974ce04e951262c4c093ce8d8a149cR94

I guess in this case since it's a BigInt, we'd make the new transform handle it. I'd say the transform.

littledan: Oh that makes sense; I didn't realize this was typically handled in a separate transform.

Does it make sense to have the 'value' field at all, though? Why bother, when you have the rawValue?

hzoo: Not sure, I guess most of the time value works for all `Literals` depending on what you want via Boolean, Null, String, etc. I guess for this it's the same thing and can't actually be represented as a number

I think it's fine to include it in the BigInt transform itself if es2015-literals are only going to handle binary, octal, etc for regular numbers.

wdhorton:
just added @mathiasbynens suggestions

wdhorton:
@hzoo @littledan is there anything keeping this from getting merged now?

hzoo:
Yeah we can iterate when we get to the transform part if it doesn't fit, tests wise it's fine to me!

hzoo:
Thanks @wdhorton for the PR, let us know if you want help with the transform as well

wdhorton:
@hzoo i have one starting question on the transform—is it ok to pull in an external lib to handle the math operations? idk what babel's policy is on adding 3rd-party dependencies, but it seems unavoidable in this case

existentialism:
@wdhorton shouldn't be an issue (guess it depends on the dep?)... we pull in `regexpu` for `transform-es2015-unicode-regex` for instance.

hzoo:
Yeah it needs a runtime library like described in https://github.com/babel/babylon/issues/569. It's basically the syntax wrapper -> those calls to the library. It's similar to transform-runtime/regenerator in that way

PR #589, opened 2017-06-23
Add delete check and fix nested class parsing for private fields

Qantas94Heavy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | N/A
| License           | MIT

This PR adds a check to make sure the `delete` operator is not applied to a private field ([spec][1]).

This also fixes an issue where private fields fail to be parsed when a method within the class contains a nested class.

[1]: https://tc39.github.io/proposal-private-fields/#sec-delete-operator-static-semantics-early-errors

### DIFF
--- a/src/parser/expression.js
+++ b/src/parser/expression.js
@@ -252,8 +252,16 @@ export default class ExpressionParser extends LValParser {
 
       if (update) {
         this.checkLVal(node.argument, undefined, undefined, "prefix operation");
-      } else if (this.state.strict && node.operator === "delete" && node.argument.type === "Identifier") {
-        this.raise(node.start, "Deleting local variable in strict mode");
+      } else if (this.state.strict && node.operator === "delete") {
+        const arg = node.argument;
+
+        if (arg.type === "Identifier") {
+          this.raise(node.start, "Deleting local variable in strict mode");
+        } else if (this.hasPlugin("classPrivateProperties")) {
+          if (arg.type === "PrivateName" || (arg.type === "MemberExpression" && arg.property.type === "PrivateName")) {
+            this.raise(node.start, "Deleting a private field is not allowed");
+          }
+        }
       }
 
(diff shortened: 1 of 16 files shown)


src/parser/expression.js
```diff
@@ -252,8 +252,16 @@ export default class ExpressionParser extends LValParser {
...
         this.checkLVal(node.argument, undefined, undefined, "prefix operation");
-      } else if (this.state.strict && node.operator === "delete" && node.argument.type === "Identifier") {
-        this.raise(node.start, "Deleting local variable in strict mode");
+      } else if (this.state.strict && node.operator === "delete") {
+        const arg = node.argument;
+
+        if (arg.type === "Identifier") {
+          this.raise(node.start, "Deleting local variable in strict mode");
+        } else if (this.hasPlugin("classPrivateProperties")) {
+          if (arg.type === "PrivateName" || arg.property.type === "PrivateName") {
```
Kovensky: Is the argument always guaranteed to be `Identifier`, `PrivateName` or `MemberExpression`?

If it happens to be neither `PrivateName` or `MemberExpression` in this branch it could throw `TypeError`.

Qantas94Heavy: Right -- I'll add a check for `MemberExpression`.

test/fixtures/experimental/class-private-properties/nested/actual.js
```diff
@@ -0,0 +1,40 @@
+class Point {
+    #x = 1;
+    #y = 2;
+
+    constructor(x = 0, y = 0) {
+        #x = +x;
+        #y = +y;
+
+
+        this.foo = class {
```
hzoo: I'm already like why would you ever do this 🙂  but ok.

PR #590, opened 2017-06-23
Follow-up on Decorators PR

### LINKED ISSUE #524: Multiple Decorators with Class as Key
From https://github.com/prettier/prettier/issues/1591:

```
@node
@view({
  store: class Foo {  
  }
})
class El {
  
}
```

Produces:
http://astexplorer.net/#/gist/0a081643680cce29b7bc6363c066a440/d38ca7ae3f503e1d19200039391831f3afd9d70c

peey:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #524, #14
| License           | MIT 

1. [commit 1](https://github.com/babel/babylon/commit/add8e4ad13b4a44de409e283e99679a904012317) Add a more helpful error message for when `@dec export class {}` is encountered, suggesting the user to use `export @dec class {}` instead (only happens in the new decorators2 plugin). [As requested](https://github.com/babel/babylon/pull/587#pullrequestreview-45734043) by @hzoo
2. [commit 2](https://github.com/babel/babylon/commit/37fa77e84f4ebe8f7fd6ea1fc3e79eae875652db) Fixes #524 by using a stack of decorators to allow nesting instead of using one global list of decorators. Earlier the following code would have attached both `@outer` and `@inner` to Foo but now it works correctly
   ```js
   @outer(@inner class Foo {})
   class Bar {}
   ```
3. [commit 3](https://github.com/babel/babylon/commit/2c8fc756434a7980d23ddcbff5347ad57ef30884) adds the only useful test case I found in #14. Rest of tests are either already present or refer to an outdated version of the spec. It can now be closed.

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -173,6 +173,11 @@ export default class StatementParser extends ExpressionParser {
       return;
     }
 
+    // special error for the common case of @dec export class
+    if (!allowExport && this.match(tt._export)) {
+      this.raise(this.state.start, "Using the export keyword between decorators and class is disallowed. Please use `export @dec class` instead");
+    }
+
     if (!this.match(tt._class)) {
       this.raise(this.state.start, "Leading decorators must be attached to a class declaration");
     }
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -177,7 +177,7 @@ export default class StatementParser extends ExpressionParser {
 
     // special error for the common case of @dec export class
     if (!allowExport && this.match(tt._export)) {
-      this.raise(this.state.start, "Using the export keyword between decorators and class is disallowed. Please use `export @dec class` instead");
+      this.raise(this.state.start, "Using the export keyword between a decorator and a class is not allowed. Please use `export @dec class` instead");
     }
 
     if (!this.match(tt._class)) {
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -171,13 +171,12 @@ export default class StatementParser extends ExpressionParser {
       currentContextDecorators.push(decorator);
     }
 
-    if (allowExport && this.match(tt._export)) {
-      return;
-    }
-
-    // special error for the common case of @dec export class
-    if (!allowExport && this.match(tt._export)) {
-      this.raise(this.state.start, "Using the export keyword between a decorator and a class is not allowed. Please use `export @dec class` instead");
+    if (this.match(tt._export)) {
+      if (allowExport) {
+        return;
+      } else {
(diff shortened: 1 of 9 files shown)


src/parser/statement.js
```diff
@@ -173,6 +173,11 @@ export default class StatementParser extends ExpressionParser {
       return;
     }
 
+    // special error for the common case of @dec export class
+    if (!allowExport && this.match(tt._export)) {
+      this.raise(this.state.start, "Using the export keyword between decorators and class is disallowed. Please use `export @dec class` instead");
```
existentialism: Nit, might read a little better as:

> Using the export keyword between a decorator and a class is not allowed.  Please use `export @dec class` instead"

peey: Yep, I think that's better too. Will make changes and push it

src/parser/statement.js
```diff
@@ -164,15 +165,21 @@ export default class StatementParser extends ExpressionParser {
...
 
     if (allowExport && this.match(tt._export)) {
       return;
     }
 
+    // special error for the common case of @dec export class
+    if (!allowExport && this.match(tt._export)) {
+      this.raise(this.state.start, "Using the export keyword between a decorator and a class is not allowed. Please use `export @dec class` instead");
+    }
+
```
bakkot: I'd combine this with the above clause, personally - 

```js
if (this.match(tt._export)) {
  if (allowExport) {
    return;
  }
  this.raise(this.state.start, "Using the export keyword between a decorator and a class is not allowed. Please use `export @dec class` instead");
}
```

src/tokenizer/state.js
```diff
@@ -89,8 +89,9 @@ export default class State {
   // Labels in scope.
   labels: Array<{ kind: ?("loop" | "switch"), statementStart?: number }>;
 
-  // Leading decorators.
-  decorators: Array<N.Decorator>;
+  // Leading decorators. Last element of the stack represents the decorators in current context.
+  // Supports nesting of decorators, e.g. @foo(@bar class {}) class {}
```
bakkot: I'd clarify here that `foo` applies only to the latter class and `bar` only to the inner.

bakkot:
Haven't checked over all the output carefully, especially the source locations, but otherwise LGTM.

peey:
@hzoo do note that this isn't as big a change as the previous PR was so it wouldn't need as many reviews (in fact we could do with 2-3 reviews just like for any standard bugfix PR).

hzoo:
Yeah I don't think we need all the reviews just pinging people anyway

PR #596, opened 2017-06-28
Distinguish between ternary's : and arrow fn's return type

nicolo-ribaudo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | #58, #429  
| License           | MIT

@hzoo This is #573 but without flow errors. Apart from 5f176cbb3dc0fe4bbbda47b0698e89f9d61c5630, this code has already been reviewed.


nicolo-ribaudo:
I rebased to resolve the merge conflicts. I squashed the commits because, since I wrote an implementation of the logic to avoid throwing errors for parenthesized expressions which are not valid parameters and then totally rewrote it, I would have needed to resolve the conflicts for a lot of unuseful commits.

The original commits can be found at https://github.com/nicolo-ribaudo/babylon/tree/issue-58-ternary-arrow-flow-copy

nicolo-ribaudo:
I just pushed some minor changes (fixed three new flow errors and removed some duplicated code)

PR #597, opened 2017-06-28
Remove future plans from README

kittens:
Right now the claims made aren't true and they just add confusion. "Interspatial parsing" was meant to refer to parsing a file when it changes and only parsing the change and then placing it back into the AST, but I think I just came up with that name, there's probably an actual name for it.

cc @hzoo


hzoo:
Yep sounds good to me!

hzoo:
> "Interspatial parsing"

That does sound pretty interesting to me, and I guess maybe that can be related to the whole immutable ast and caching that stuff?

PR #608, opened 2017-07-02
ClassFields: Allowing comma separated grammar

### LINKED ISSUE #540: Class Fields (combined) - Stage 2
> Updates on class-field proposals (both public and private)

Champions: @jeffmo (public class fields) @littledan (private + combined)
Spec Repo: https://github.com/littledan/proposal-class-fields/
Spec Text: https://littledan.github.io/proposal-class-fields/
Slides: https://drive.google.com/file/d/0B-TAClBGyqSxWHpyYmg2UnRHc28/view

> Remember that proposals are just that; subject to change until Stage 4!

## Examples

```js
class C {
  static x = 1, #y, [a];
  z, #w = 2, [b];
  
  a() {
    this.z;
    this.#w++;
    #w; // #w is this.#w
  }
}
```

## Parsing/ESTree
- [x] AST handled already in current `classProperties` and `classPrivateProperties`.
- [ ] May need to rename though (new plugin name).
- [ ] make sure no "private computed" parse
- [ ] need to support comma separated properties.

## Transform
- [ ] combine the class-properties plugin + tbd private properties one

## Contacts
- [#proposal-class-fields](https://babeljs.slack.com/messages/proposal-class-fields) on Slack!
- @littledan 
- @diervo (dval on Slack)
- @Qantas94Heavy

diervo:
Refactor for classFields:
- Fixes #540
- Fixes location for publicFields (should not contain semi-colon)
- Allow decorators for comma separated fields
- Added tests

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | yes
| Deprecations?     | yes
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #540
| License           | MIT


src/parser/statement.js
```diff
@@ -881,10 +884,26 @@ export default class StatementParser extends ExpressionParser {
...
 
       this.parseClassMember(classBody, member, state);
+      const isClassField = member.kind == null;
+
+      if (isClassField) {
+        if (this.eat(tt.comma)) {
+          this.state.inCommaSeparatedFields = true;
+          // Copy decorators for cases like `@foo x, y;`
+          if (decorators.length) {
+            member.decorators = decorators;
```
jridgewell: Let's slice here. It'll be really weird if a transform pushes a decorator and two fields get it.

src/parser/statement.js
```diff
@@ -899,7 +918,7 @@ export default class StatementParser extends ExpressionParser {
       }
     }
 
-    if (decorators.length) {
+    if (!this.state.inCommaSeparatedFields && decorators.length) {
```
jridgewell: I don't understand why `inCommaSeparatedFields` affects this. Maybe it should be a separate check instead (if fields can't have a trailing comma)?

src/parser/statement.js
```diff
@@ -1071,7 +1090,7 @@ export default class StatementParser extends ExpressionParser {
         /* isConstructor */ false,
       );
       this.checkGetterSetterParamCount(method);
-    } else if (this.isLineTerminator()) {
+    } else if (this.isLineTerminator() || this.match(tt.comma)) {
```
jridgewell: Does this mean methods can have trailing commas?

```js
class Foo {
  method() {
  },
}
```

I don't see a test for it anywhere.

src/parser/statement.js
```diff
@@ -881,10 +884,26 @@ export default class StatementParser extends ExpressionParser {
...
         }
-        decorators = [];
       }
 
       this.parseClassMember(classBody, member, state);
+      const isClassField = member.kind == null;
+
+      if (isClassField) {
+        if (this.eat(tt.comma)) {
+          this.state.inCommaSeparatedFields = true;
```
jridgewell: We can probably avoid this state entirely.

nicolo-ribaudo:
> Fixes location for publicFields (should not contain semi-colon)

Why? Usually the semicolon is included:

```js
var a = 2; // start: 0, end: 10
```
```js
foo(); // start: 0, end: 6
```

src/parser/statement.js
diervo: Good call.

diervo:
@nicolo-ribaudo Not for comma separated statements, which is what I'm refering here.
```js
import { a,b,c } from "x"

var x = 1, b =2, z = 3;
```
Will try to to make it consistent.

src/parser/statement.js
diervo: Will add more tests.

src/parser/statement.js
diervo: I don't think we can completely get away without knowing we are within the context of a coma separated production due to ASI and explicit checks for ;.
Im revising it nonetheless.

diervo:
Ok based on the comments I refactor pretty much the whole thing:
- Added more tests
- All locations should be correct now (@nicolo-ribaudo)
- Refactor for better parse errors when we a re in the comma separated case.

@jridgewell Due to the abstraction we have where we call `parseClassMember`, we need to pass in the context in order to disambiguate if we are in a comma separated statement to throw unexpected characters when necessary.

nicolo-ribaudo:
I have a doubt about decorators:

```js
class A {
  @dec x, y;
}
class B {
  @dec x;
  y;
}
class C {
  @dec x;
  @dec y;
}
```

Is `A` more like `B` or like `C`? According to this pr, it is like `C`; Typescript interprets it like `B`.
The `decorators2` plugin doesn't have this problem because it disallows decorators on class prpoerties, but we should decide how must `decrators` behave,

diervo:
`A` and `C` are equivalent.
Decorators should be applied to all the comma separated fields (keep me honest @littledan)

@peey should revisit this logic once this PR lands in master.

littledan:
@diervo Agree with what you're saying about equivalency (except it matters how many times `dec` is evaluated--imagine it's a getter on the global object, or inside a `with`). A decorator applying to multiple comma-separated elements is the entire reason comma-separated elements were proposed by @wycats; maybe he can provide more clarification.

Some test ideas:
- Static fields that are comma-separated: `class C { static x, y }`
- Invalid to put a comma between two methods `class C { a(){}, b() {} }`
- Mixed public and private fields that are comma-separated `class C { #x, y }`

cc @bakkot

nicolo-ribaudo:
> except it matters how many times dec is evaluated

Then the AST must somehow reflect the difference between a decorator applied to a list of properties and the same decorator applied to different properties. Otherwise how could `babel-generator`/`prettier`/`any other code generator` know what code to emit?

---

e.g.
<details>
<summary>AST</summary>

```js
Program {
  body: [
    ClassDeclaration {
      id: Identifier { name: "Foo" },
      body: ClassBody {
        body: [
          ClassProperty {
            decorators: [
              Decorator {
                expression: CallExpression {
                  callee: Identifier { name: "dec" },
                  arguments: []
                }
              }
            ],
            key: Identifier { name: "a" }
          },
          ClassProperty {
            decorators: [
              Decorator {
                expression: CallExpression {
                  callee: Identifier { name: "dec" },
                  arguments: []
                }
              }
            ],
            key: Identifier { name: "b" }
          }
        ],
      },
    },
    FunctionDeclaration {
      params: [],
      body: BlockStatement {
        body: [
          "dec = () => { throw new Error(); }",
          "return /* super cool decorator */"
        ]
      }
    }
  ]
}
```
</details>

This ast represents to different programs: one works and the other throws.
```js
class Foo {
  @dec() a, b; // works
}

function dec() {
  dec = () => { throw new Error(); };
  return decorator;
}
```
```js
class Foo {
  @dec() a;
  @dec() b; // Throws
}

function dec() {
  dec = () => { throw new Error(); };
  return decorator;
}
```

---

A possible solution is to represent a comma-separated list of properties actually as a list of properties, similar to a `VariableDeclaration`:

```js
class Foo {
  @dec() a, b;
}
```

<details>
<summary>AST</summary>

```js
ClassDeclaration {
  id: Identifier { name: "Foo" },
  body: ClassBody {
    body: [
      ClassPropertiesList {
        decorators: [
          Decorator {
            expression: CallExpression {
              callee: Identifier { name: "dec" },
              arguments: []
            }
          }
        ],
        static: false,
        properties: [
          ClassProperty {
            key: Identifier { name: "a" }
          },
          ClassProperty {
            key: Identifier { name: "b" }
          }
        ],
      }
    ],
  },
}
```

diervo:
@littledan @nicolo-ribaudo I think you are both right, but this will require some interesting refactor, and probably a somewhat big breaking change on the transformers for publicFields.

Is everyone ok then moving to the following productions (analogous to `VariableDeclaration`):

`FieldDefinition`:
  +decorators: []
  +definitions: [`PublicField` | `PrivateField`]

Basically any public/private fields will be enclosed by FieldDefinition.

If we do this, we might as well refactor the whole ClassBody to make it future proof as @jridgewell was suggested to do in #609.

If you give me thumbs up I can give it a run today, otherwise will have to wait til next weekend most likely or someone else can take over.

Thoughts?

nicolo-ribaudo:
> and probably a somewhat big breaking change

In #609 @xtuc suggested to merge the `classProperties` and `classPrivateProperties`. We can enable the changes introduced by this pr only under the new plugin to avoid breaking changes

diervo:
@nicolo-ribaudo The problem is not creating a new plugin, is the fact that we are changing the shape of the Nodes (for example no longer ClassProperty will have decorators, or potentially a change on the node name ClassProperty =>PublicField ) that transformers expect.

And I'm very reluctant to have a fork in the code that generate different Node types based on the syntax plugin that is enabled, specially since we are working in babel7 where we have more room for error. If we do this, my vote is to do non backward compatibility change.

@hzoo what you think?

littledan:
If this is for Babel 7, could the legacy decorator syntax and transform also be updated to the new AST as well? It seems like the new form you're proposing is simply more general. Are there a lot of transforms out there which depend on the current shape of decorators?

diervo:
Closing this since comma separated proposal was dropped from the original classField proposal and will be advancing separately.

littledan:
If someone wants to reopen the PR and pick it up, you can find the new proposal at https://github.com/littledan/proposal-comma-separated-fields . The default class fields flag needs to prohibit comma separation, with a separate flag to permit it; I am not sure whether this flag behavior should be implemented at the babylon or transform level.

PR #609, opened 2017-07-02
Add static private class field support

jridgewell:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

Private class fields can be static (according to the [unified proposal](https://github.com/tc39/proposal-class-fields)).

### DIFF
--- a/test/fixtures/experimental/class-private-properties/static/actual.js
+++ b/test/fixtures/experimental/class-private-properties/static/actual.js
@@ -1,0 +1,4 @@
+class A {
+  static #x;
+  static #y = 1;
+}
--- a/ast/spec.md
+++ b/ast/spec.md
@@ -1056,6 +1056,7 @@ interface ClassProperty <: Node {
   type: "ClassProperty";
   key: Expression;
   value: Expression;
+  static: boolean;
   computed: boolean;
 }
 ```
@@ -1067,6 +1068,7 @@ interface ClassPrivateProperty <: Node {
   type: "ClassPrivateProperty";
   key: Identifier;
   value: Expression;
+  static: boolean;
 }
 ```
 
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -922,15 +922,6 @@ export default class StatementParser extends ExpressionParser {
     const method: N.ClassMethod = memberAny;
     const prop: N.ClassProperty = memberAny;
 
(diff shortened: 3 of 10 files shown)


xtuc:
Nice job @jridgewell.

Do you think we could merge classProperties and classPrivateProperties plugins in Babylon? Since it just adds the static syntax.

jridgewell:
I'll do that in a second PR.

diervo:
@xtuc @jridgewell 
We should wait for merging both plugins until publicFields get to Stage3, which hopefully will happen in the next TC39 meeting (in about a month). 

There is still some debate on the committee on wether `this.#x` will be required and so let's wait for the final version.

Decorators might be another moving piece that is very intertwined, but I guess is mostly implementation details.

Lastly, just an FYI I think @littledan wants to present private methods, so let's see which Stage that lands... :)

test/fixtures/experimental/class-private-properties/static/actual.js
```diff
@@ -0,0 +1,4 @@
+class A {
+  static #x;
+  static #y = 1;
+}
```
littledan: Have comma-separated field declarations landed yet? If so, it'd be nice to have tests that demonstrate that you can mix static public and private field declarations in a comma-separated list.

littledan:
This looks great. I don't see why it should wait until Stage 3 to land--I presented static private fields to the committee in the May meeting with this syntax and didn't hear any concerns. I think @hzoo is doing a lot of work to communicate to users that Stage 2 proposals are unstable.

For decorators, we unfortunately don't have spec text yet which combines it with fields; hope to fix that soon. Private methods haven't been presented yet, so I guess they will wait a bit longer to add to Babel.

test/fixtures/experimental/class-private-properties/static/actual.js
diervo: PR is pending

diervo:
@littledan I was not talking about this PR, this is good to go, the refactor to combine public/private fields is what Im saying to postpone til we figure out the `this.#x` syntax.

Issue #289, opened 2017-01-12
Incorrect parsing of escaped DirectiveLiteral string


vjeux:
Input

```js
'\''
```

Babylon 6 value (incorrect) http://astexplorer.net/#/heginnP81D

![image](https://cloud.githubusercontent.com/assets/197597/21877651/54e5cecc-d841-11e6-91a7-0186bf4d1378.png)

Flow parser value (correct) http://astexplorer.net/#/heginnP81D/2

![image](https://cloud.githubusercontent.com/assets/197597/21877669/6dadef8e-d841-11e6-8119-96250cf68fe6.png)

vjeux:
cc @hzoo

DrewML:
Not saying this isn't an issue, but for the sake of prioritizing, is this something that has been hit in real code, or just a problem you spotted?

vjeux:
I hit this in a unit test on prettier ( https://github.com/jlongster/prettier/blob/master/tests/strings/strings.js#L4 ). I don't expect it to cause trouble in real code.

danez:
I created a PR for this, but only when using the `estree` plugin. In "normal" babel-mode the value of the `DirectiveLiteral.value` is considered to be the raw value without quotes.

Issue #216, opened 2016-11-09
String literal parsed incorrectly when containing "\\"


jeanlauliac:
### Input Code

To reproduce:

1. Get babylon 6.13.1
2. Start node REPL
3. Run these lines:

```js
b = require('babylon');
b.parse('("C:\\root\\arbitrary.js")').program.body[0].expression;
```

Here is the result:

```
Node {
  type: 'StringLiteral',
  start: 1,
  end: 23,
  loc:
   SourceLocation {
     start: Position { line: 1, column: 1 },
     end: Position { line: 1, column: 23 } },
  extra:
   { rawValue: 'C:\rootarbitrary.js',
     raw: '"C:\\root\\arbitrary.js"',
     parenthesized: true,
     parenStart: 0 },
  value: 'C:\rootarbitrary.js' }
```

I believe this is incorrect. Look at the equivalent "value" babylon provides. It is `'C:\rootarbitrary.js'`, but I believe it should be `'C:\root\arbitrary.js'`. Basically it seems babylon parses the backslashes incorrectly under some conditions.

### Babylon/Babel Configuration

No particular config, just call Babylon from the node.js REPL.

### Your Environment

| software         | version
| ---------------- | -------
| Babylon          |  6.13.1
| node             | v5.0.0
| npm              |  none
| Operating System |  OS X 10.11.6

jeanlauliac:
Alright, I realised my mistake, this is not a bug. This is because we have a string embedded within another, so we need to double-escape.

PR #616, opened 2017-07-10
Directive value should be consistent with strings.

ashsearle:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | none added, tests continue to pass (and so do babel tests)
| Fixed tickets     | see below
| License           | MIT

Change the implementation of directive so the `value` is consistent with strings.  Match this with internal change to use `raw` value of directives to maintain spec-compliance.

Why?  Because the current implementation causes unwanted confusion as seen in issues #216 and #289.

When this change works its way downstream it will fix https://github.com/babel/babel/issues/4767, https://github.com/babel/babili/issues/308 and https://github.com/babel/babili/issues/308 and ensure developers can work with minimal test-cases for new string-related features such as https://github.com/babel/babili/issues/619.


danez:
I'm not sure changing this is the way to go, as it might break a lot of plugins who expect `value` to be the raw value.

I think the definition of Directive was always that `value` is the raw value without quotes and `raw` the raw value with quotes. 

We could also change the way Directives are handled and revert back to the estree way. That of course is a similar breaking change.

Looking at the related issues, it seems that maybe some plugins are handling DirectiveLiteral the same way as StringLiteral, which it right now clearly shouldn't.

ashsearle:
@danez It's not just plugins that have problems, there's an issue in `babel-generator` which currently handle directives as though they're normal string literals.  Please see https://github.com/babel/babel/issues/4767 (and pull request https://github.com/babel/babel/pull/5935)

danez:

Thanks for your effort.

PR #626, opened 2017-07-13
Fix parsing a <!-- b in modules

### LINKED ISSUE #622: `a <!-- b` in modules are actually equivalent to `a < !(--b)`
See: https://github.com/tc39/ecma262/issues/949

An error is generated for `a <!-- b` with `scriptType: "module"`, when it should parse like `a < !(--b)`.

existentialism:
Fixes #622.


test/fixtures/typescript/function/expression/expected.json
```diff
@@ -0,0 +1,33 @@
+{
```
jridgewell: What's this?

PR #632, opened 2017-07-15
Position obj perf

hzoo:
returning an object instead of calling a function?

https://github.com/babel/babylon_performance but getting mixed results so might be doing something wrong

### DIFF
--- a/src/parser/node.js
+++ b/src/parser/node.js
@@ -2,7 +2,7 @@
 
 import Parser from "./index";
 import UtilParser from "./util";
-import { SourceLocation, type Position } from "../util/location";
+import type { SourceLocation, Position } from "../util/location";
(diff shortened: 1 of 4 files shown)


src/parser/node.js
```diff
@@ -14,8 +14,13 @@ class Node implements NodeBase {
     this.type = "";
     this.start = pos;
     this.end = 0;
-    this.loc = new SourceLocation(loc);
```
xtuc: I would expect a function returning an object to be as fast as just an object (since it will inline it quickly by the engine). Could we try with a function `MakeSourceLocation` which returns the location object.

Otherwise ok for me :+1:

hzoo: ^ in loose mode it's going to compile the class into a function anyway. @jridgewell said it's not worth it and it seems to not big that big a when anyway and to focus on the object shape instead

PR #634, opened 2017-07-16
Add optionality to catch bindings

MarckK:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | 
| Spec compliancy?  | 
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

This implements the optional catch binding proposal (babel/proposals [#7](https://github.com/babel/proposals/issues/7)).


bakkot:
1) I think you want to update the [AST specification](https://github.com/babel/babylon/blob/master/ast/spec.md#trystatement) too.

2) Maybe add a test for try + catch + finally, e.g. `try {} catch {} finally {}`.

Otherwise looks good to me!

src/parser/statement.js
```diff
@@ -491,12 +491,12 @@ export default class StatementParser extends ExpressionParser {
...
-      clause.param = this.parseBindingAtom();
-      this.checkLVal(clause.param, true, Object.create(null), "catch clause");
-      this.expect(tt.parenR);
-
+      if (this.match(tt.parenL) || !this.hasPlugin("optionalCatchBinding")) {
+        this.expect(tt.parenL);
+        clause.param = this.parseBindingAtom();
+        this.checkLVal(clause.param, true, Object.create(null), "catch clause");
+        this.expect(tt.parenR);
+      }
```
jridgewell: Let's set `clause.param = null` in an else case.

hzoo:
https://github.com/babel/babylon/blob/master/CONTRIBUTING.md#creating-a-new-plugin-spec-new

Yeah I would add an entry to the readme, update ast/spec

src/parser/statement.js
```diff
@@ -491,12 +491,13 @@ export default class StatementParser extends ExpressionParser {
...
+      clause.param = null;
```
hzoo: I think you'll need to add this to `else` otherwise it would set it to null in all cases? And then rerun the tests fixtures

MarckK: Hmmm.  This is causing a failure in one of my tests.  I'll have to look at more.

src/parser/statement.js
MarckK: Yes.  👍

src/parser/statement.js
```diff
@@ -491,12 +491,14 @@ export default class StatementParser extends ExpressionParser {
     if (this.match(tt._catch)) {
       const clause = this.startNode();
       this.next();
-
-      this.expect(tt.parenL);
-      clause.param = this.parseBindingAtom();
-      this.checkLVal(clause.param, true, Object.create(null), "catch clause");
-      this.expect(tt.parenR);
-
+      if (this.match(tt.parenL) || !this.hasPlugin("optionalCatchBinding")) {
```
hzoo: i was wondering whether this `this.match` check is necessary?

hzoo: Ok never mind, talked to Kara and it makes sense. This is for the case you have the plugin on, but you choose to still use the binding itself.

hzoo:
Nice work! 😄

PR #635, opened 2017-07-18
WIP Pattern Matching (Stage 0)

krzkaczor:
Work in progress. I am creating this PR so other people can see progress and contribute. Related issues: https://github.com/babel/proposals/issues/6

I will align to all contributing guidelines in a very near future ;) 

What's missing:
 - [x] solve ambiguity between pattern matching and calling a `match` function (currently match is considered just another keyword which is obviously far from perfect)
 - [ ] ast nodes docs
 - [ ] some patterns ( ex. `[...]`) doesnt work
 - [ ] test cases for errors
 - [ ] align with contribution guidelines

### DIFF
--- a/src/util/identifier.js
+++ b/src/util/identifier.js
@@ -31,7 +31,7 @@ export const reservedWords = {
 // And the keywords
 
 export const isKeyword = makePredicate(
-  "break case catch continue debugger default do else finally for function if return switch throw try var while with null true false instanceof typeof void delete new in this let const class extends export import yield super",
+  "break case catch continue debugger default do else finally for function if return switch throw try var while with null true false instanceof typeof void delete new in this let const class extends export import yield super match",
 );
 
 // ## Character categories
--- a/src/util/identifier.js
+++ b/src/util/identifier.js
@@ -31,7 +31,7 @@ export const reservedWords = {
 // And the keywords
 
 export const isKeyword = makePredicate(
-  "break case catch continue debugger default do else finally for function if return switch throw try var while with null true false instanceof typeof void delete new in this let const class extends export import yield super match",
+  "break case catch continue debugger default do else finally for function if return switch throw try var while with null true false instanceof typeof void delete new in this let const class extends export import yield super",
 );
 
 // ## Character categories
--- a/ast/spec.md
+++ b/ast/spec.md
@@ -78,6 +78,7 @@ These are the core Babylon AST node types.
   - [NewExpression](#newexpression)
   - [SequenceExpression](#sequenceexpression)
   - [DoExpression](#doexpression)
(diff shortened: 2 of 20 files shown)


src/util/identifier.js
```diff
@@ -31,7 +31,7 @@ export const reservedWords = {
 // And the keywords
 
 export const isKeyword = makePredicate(
-  "break case catch continue debugger default do else finally for function if return switch throw try var while with null true false instanceof typeof void delete new in this let const class extends export import yield super",
+  "break case catch continue debugger default do else finally for function if return switch throw try var while with null true false instanceof typeof void delete new in this let const class extends export import yield super match",
```
hzoo: We'll need to somehow fork this with the hasPlugin check as well right?

hzoo:
We should start a discussion on the AST structure then (spec.md change)

src/util/identifier.js
xtuc: You mean `makePredicate([...] + this.hasPlugin('match') ? 'match' : '')` ? Yes because `match` shoudln't be a keyword without the plugin.

krzkaczor: @hzoo @xtuc I wonder if we should make `match` a keyword at all (at least part of this array)? 
If we do it then we need to tweak for all situations where match is currently being used as a valid identifier.

What I imagined instead as a simplest implementation, is to tokenize it as a identifier and use lookup to determine if it should be a keyword in a given context (part of match expression or just a ex. function call `match(something);`).

xtuc: Yes, that makes sense. Or consider it as a keyword only in a certain context, but more `if`s would be spread across the codebase I guess.

xtuc:
@hzoo what a about an AST repo with the Babel AST spec? We could PR and discuss there.

I want to avoid mixing AST and implementation discussions.

hzoo:
I don't think we need an AST repo, babylon is already that repo at the very least 😛. Or just discuss in the proposal issue, I think it's fine, otherwise it would be an estree issue.

src/util/identifier.js
krzkaczor: @hzoo @xtuc this is fixed by 5857eb3

alexeyraspopov:
What’s the status of this PR? What’s left to be done? Any help needed?

krzkaczor:
@alexeyraspopov There is a small todo list in the ticket description. Unfortunately, I don't have time right now to continue work on this issue but I will come back to it in some time. 

One of the first steps now should be re-aligning it with spec proposal - I am sure that it developed further.

Feel free to continue from where I left off ;)

danez:

Thanks for your effort.

PR #637, opened 2017-07-19
Flatten TemplateElement AST

jridgewell:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     |
| License           | MIT

Time for a breaking change?

### DIFF
--- a/ast/spec.md
+++ b/ast/spec.md
@@ -962,10 +962,8 @@ interface TaggedTemplateExpression <: Expression {
 interface TemplateElement <: Node {
   type: "TemplateElement";
   tail: boolean;
-  value: {
-    cooked: string | null;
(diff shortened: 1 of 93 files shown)


danez:
What is the intention behind this? I would rather first open an issue at estree if you want to change this, but I'm personally strongly against diverging from estree any more.

hzoo:
Yeah we already felt the issues with diverging earlier, so would rather not unless we decide to completely do so in v8. Basically it's not a lot of benefit to change 1 thing in the AST at the end of releasing v7

PR #643, opened 2017-07-20
Flow opaque type aliases parsing

jbrown215:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | N/A
| License           | MIT

Flow v0.49 added parser support for opaque type aliases. Recently, in [this commit](https://github.com/facebook/flow/commit/412018db5c8c2b12c6c126d8d93f8c28d3f2f152), Flow finalized the parser changes for opaque type aliases. 

The syntax is:
`(export?) opaque type alias <identifier>(: <super-type>?) = <impltype>`
where export and the supertype are optional.

If we are in a declare, we do not parse the impltype. So the syntax in a declare is:

`declare (export?) opaque type alias <identifier>(: <super-type>?)`


jbrown215:
Thanks for the quick review!

I assume I'll need to put up a similar PR against 6.x?

existentialism:
@jbrown215 yeah, can backport to 6.x after this lands. may also want to add support for `DeclareOpaqueType`, `OpaqueType`, etc to babel-types and babel-generator after this lands too.

PR #648, opened 2017-07-25
Remove private field shorthand

Qantas94Heavy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

There seems to be tentative agreement to remove the private field
shorthand given the added confusion and edge cases involved with the
shorthand.

Refs: https://github.com/tc39/proposal-class-fields/issues/21

----

This will change the expected error message for the ASI generator test, given that `#x` would no longer be a valid expression, which means the error will be thrown there instead.


littledan:
Note that there is a possible follow-on proposal to reintroduce the shorthand at https://github.com/littledan/proposal-private-shorthand

Issue #640, opened 2017-07-19
Wrong parsing of new.target inside an arrow function


droooney:
Input:

```js
const foo = () => {
  console.log(new.target);
};
```

Expected behaviour: a syntax error (Chrome throws "new.target expression is not allowed here", Firefox throws "new.target only allowed within functions").

Actual behaviour: parsed.

hzoo:
Hey @droooney! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

Issue #641, opened 2017-07-19
Wrong parsing of new.target inside function arguments


droooney:
babylon version: **6.17.4** (the same for v7).

Input:

```js
function Foo(bar = new.target) {}
```

Expected behaviour: parsed.

Actual behaviour: a syntax error is thrown.

The behaviour for arrow functions should differ though, because function arguments are inside the function body (see #640).

hzoo:
Hey @droooney! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

PR #652, opened 2017-07-28
Parse parameters inside function's env

nicolo-ribaudo:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #641
| License           | MIT

With this pr, `this.state.inFunction` is set before parsing the function parameters. This allows `new.target` to be inside function default parameters.

I didn't do the same thing with `inAsync` and `inGenerator` because, even if yield and await should be parsed as `YieldExpression` and `AwaitExpression`, they are an early error so babylon doesn't need to parse them.


danez:
Does this also fix #640?

nicolo-ribaudo:
I don't think

Issue #633, opened 2017-07-15
Incorporating Test262


jugglinmike:
[I recently extended the Test262 project to include additional metadata for "negative" tests](https://github.com/tc39/test262/pull/778). Specifically: tests that are expected to produce early errors and syntax errors can now be automatically distinguished for tests that are expected to produce runtime errors.

My goal was to make the tests consumable by JavaScript parsers. I've since incorporated the tests into [the JSHint project](https://github.com/jshint/jshint), and it's been very useful for fixing bugs and preventing regressions. I'd like to help other Free and Open Source parsers do the same.

Judging from the response to [a previously-reported issue on the topic](https://github.com/babel/babylon/issues/327), I believe there is interest for including Test262 here. That issue was closed with Babel's inclusion of the project, but I think there is still value in testing the parser itself. If any of the maintainers can confirm their interest, I would be happy to follow up with some more specific practical questions (e.g. submodule usage, etc.) and go from there.

hzoo:
Hey @jugglinmike! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

hzoo:
Hey @jugglinmike thanks for reaching out 😀! I've also talked to @leobalter/@littledan a bit about this so definetely have had some interest for a while (also ref https://twitter.com/littledan/status/880436506996994048)

> Also your previous link is referring to this issue currently

jugglinmike:
Excellent :) Some initial questions:

- How should the Test262 project be included here? Some possibilities include a
  git submodule or a git checkout that is fetched as necessary at test time
  (and excluded from the project via a `.gitignore` file)
- How should the tests be executed? My initial guess would be to define a new
  target in the project's `Makefile`, but I don't know if you have portability
  concerns with that approach
- Related to the above, how should test runs be scheduled? I imagine you'll
  want a new "job" in the TravisCI system dedicated to this. They could be run
  as part of the `npm test` script, but this will likely be a little
  heavy-handed for general-purpose development.
- Would you like these tests to run in anything other than the current LTS
  release of Node.js?

With those answered, I should be able to put together an initial proof of
concept so that we have something more concrete to discuss.

> Also your previous link is referring to this issue currently

[...]

PR #654, opened 2017-07-29
Integrate Test262

jugglinmike:
Introduce a GNU Make target for retrieving TC-39's Test262 suite and
validating parsing of the files it contains. Interpret each file as a
parser test in accordance with that project's `INTERPRETING.md`
document. Allow for the specification of allowed failures via a
"whitelist" file so that the test suite may help prevent regressions in
this project in situations where this project has known bugs. Initialize
the "whitelist" file with a listing of all tests that are currently
failing. Extend the continuous integration environment's configuration
to automatically run these tests.

| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | gh-633
| License           | MIT

Hey @hzoo!

Here's my first attempt at integrating Test262. There's a lot of new code here,
so I'm prepared to make a log of changes according to review feedback.

One thing I would prefer to do is split the `run_test262_utils.js` file into
multiple files, possibly within a dedicated directory. There's not much
precedent for that within the `scripts` directory, so I thought I'd hold off on
that until I heard back from you.

I'd also like to point out that the code within the `scripts` directory is not
currently being linted by the npm "lint" command. I manually satisfied the
linter anyway, with one exception: trailing commas. (I'm developing in Node.js
current LTS release where trailing commas are not supported.) Is there anything
you would like to change about this?


nicolo-ribaudo:
Travis reported this error:
```
$ if [ "$JOB" = "test262-test" ]; then make test-test262; fi
node scripts/run_test262.js
{ Error: ENOENT: no such file or directory, stat '/home/travis/build/babel/babylon/build/test262/test'
    at Error (native)
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/home/travis/build/babel/babylon/build/test262/test' }
```
Is it intended?

scripts/run_test262.js
```diff
@@ -0,0 +1,104 @@
...
+      console.log("");
+      console.log("Details:");
+      console.log(badnewsDetails.join("\n").replace(/^/gm, "   "));
+    }
+
+    process.exitCode = summary.passed ? 0 : 1;
+  })
+  .catch(function(err) {
+    console.error(err);
+    process.statusCode = 1;
```
nicolo-ribaudo: This should be `process.exitCode`?

scripts/test262_whitelist.txt
```diff
@@ -0,0 +1,7972 @@
+# This file lists tests that are known to produce incorrect results when parsed
+# with Babyline:
```
nicolo-ribaudo: nit: Babylon

nicolo-ribaudo:
I see there are a lot of failing tests which babylon actually supports via plugins, for example `language/expressions/async-generator/args-trailing-comma-multiple.js`.
I think there should be a way to specify which plugins enable for a given file/folder.

scripts/run_test262.js
jugglinmike: Yes. This is critical--thank you for spotting it.

jugglinmike:
> I think there should be a way to specify which plugins enable for a given
> file/folder. Maybe we should use the features flag?

I'd be open to try that, but at some point, I think we should consider
introducing a YAML parser. The ad-hoc regular expressions I've written seem
"close enough" for the current use case, but using RegExps to interpret dynamic
values like that is much more error prone. I've been trying to avoid adding any
dependencies for this patch, so I'm wondering how the project maintainers would
feel about a new "devDependency" on a YAML parser.

jugglinmike:
Alternatively, we could unconditionally enable all plugins for all tests. This is certainly less precision, but since Test262 generally doesn't include any "future hostile" tests, it may be an acceptable compromise.

hzoo:
We just did something like this for our printer tests (generator) https://github.com/babel/babel/pull/6018 with explicit flags. We could do unconditional for all tests though for now, if someone runs the stage 0 preset, it would be the same thing and none of the plugins should be incompatible.

scripts/run_test262_utils.js
```diff
@@ -0,0 +1,223 @@
...
+      );
+    })
+    .then(flatten);
+};
+
+exports.runTest = function(test) {
+  const sourceType = test.isModule ? "module" : "script";
+
+  try {
+    parse(test.content, { sourceType: sourceType });
```
hzoo: Just need to add the ones in https://github.com/babel/babylon#plugins

hzoo:
is there already a script/way to remove the whitelist tests automatically if things are fixed later?

scripts/run_test262.js
```diff
@@ -0,0 +1,104 @@
...
+
+const path = require("path");
+const chalk = require("chalk");
+const utils = require("./run_test262_utils");
+
+const testDir = path.join(__dirname, "..", "build", "test262", "test");
+const whitelistFile = path.join(__dirname, "test262_whitelist.txt");
+
+Promise.all([utils.getTests(testDir), utils.getWhitelist(whitelistFile)])
+  .then(function([tests, whitelist]) {
```
Kovensky: This breaks in Node 4 (do we care?)

scripts/run_test262.js
```diff
@@ -0,0 +1,104 @@
...
+        " invalid programs did not produce a parsing error" +
+        " (and allowed by the whitelist file)",
+      summary.allowed.falseNegative.length +
+        " valid programs produced a parsing error" +
+        " (and allowed by the whitelist file)",
+    ];
+    const badnews = [];
+    const badnewsDetails = [];
+
+    [
```
Kovensky: Lines starting with a `[` are pretty scary, regardless of semicolons.

Maybe save this into a `const`, or make it the right side of a `for of`?

scripts/run_test262.js
```diff
@@ -0,0 +1,104 @@
...
+      if (!tests.length) {
+        return;
+      }
+
+      const desc = tests.length + " " + label;
+
+      badnews.push(desc);
+      badnewsDetails.push(desc + ":");
+      badnewsDetails.push(
+        ...tests.map(function(test) {
```
Kovensky: This also breaks in Node 4 😄

scripts/run_test262_utils.js
```diff
@@ -0,0 +1,223 @@
...
+
+function hasEarlyError(src) {
+  return !!(
+    src.match(/^\s*negative:\s*$/m) && src.match(/^\s+phase:\s*early\s*$/m)
+  );
+}
+
+function readDir(dirName) {
+  return new Promise(function(resolve, reject) {
+    fs.readdir(dirName, function(err, contents) {
```
Kovensky: Any objections to adding a devDependency on [`util.promisify`](https://www.npmjs.com/package/util.promisify)?

It's available natively in Node >= 8, the package just shims it if it's missing.

scripts/run_test262_utils.js
```diff
@@ -0,0 +1,223 @@
...
+  const sourceType = test.isModule ? "module" : "script";
+
+  try {
+    parse(test.content, { sourceType: sourceType });
+    test.actualError = false;
+  } catch (err) {
+    test.actualError = true;
+  }
+
+  test.result = test.expectedError ^ test.actualError ? "fail" : "pass";
```
Kovensky: Better to use `!==`.

scripts/run_test262_utils.js
```diff
@@ -0,0 +1,223 @@
...
+
+exports.getWhitelist = function(filename) {
+  return readFile(filename).then(function(contents) {
+    return contents
+      .split("\n")
+      .map(function(line) {
+        return line.replace(/#.*$/, "").trim();
+      })
+      .filter(function(line) {
+        return line.length > 0;
```
Kovensky: `.filter(Boolean)` would work too but I guess this is clearer

scripts/run_test262.js
hzoo: I think it's ok to only run on latest for these?

scripts/run_test262.js
jugglinmike: For consistency, I'd prefer to keep a functional style. But I see your point. I've made this the right-hand side of a `void` expression--that will guard against ASI-related bugs, and it should also more clearly indicate to the reader that this literal value is not used beyond this statement.

scripts/run_test262_utils.js
jugglinmike: I wouldn't object, but I believe this question was @hzoo, so I'm going to hold off unless he agrees.

jugglinmike:
> is there already a script/way to remove the whitelist tests automatically if
> things are fixed later?

There was not, but that's a good idea. Maintaining the white list in JSHint has
been a pain point for me. I've updated the patch to include a command-line
flag and dedicated Makefile target.

So I think that's everything--this should be ready for another round of review.

scripts/run_test262_utils.js
hzoo: I don't think we need to run on anything other than latest node (8)? We can add it too if we really want to but I think it's fine to run one one version of node like we do for coverage and lint

jugglinmike: Works for me!

jugglinmike:
I've updated the job to run in the latest release of Node.js, and I've taken @Kovensky's advice regarding the "util.promisify" module. My thinking is that many contributors likely prefer to develop in the LTS release. Even though they may not all care to run this particular test suite, the module's dependency tree is light enough to be justify the convenience.

The `babel-test` job is failing. Is there something we should do about that in this patch?

hzoo:
> There was not, but that's a good idea. 

Like it would basically run the whitelist tests and if they pass correctly instead of erroring/not erroring it could tell you to remove them (or it is done automatically), similar to updating a snapshot/fixture

hzoo:
oops I just merged https://github.com/babel/babel/pull/6056, which means we need to update yarn at least?

jugglinmike:
Whoops; before you updated the comment, I thought you were saying I should update the yarn lock file. I guess wasn't necessary, after all (I'm not familiar with that tool). Should I revert this latest commit?

hzoo:
I updated Babel to use a new feature in yarn; I think we have to add https://github.com/babel/babel/pull/6056/files#diff-354f30a63fb0907d4ad57269548329e3L15 since the yarn version it's using now is older. Not sure it's going to fix but it's an issue due to us targeting master. Either way it would of broke in someone else's PR

hzoo:
Yeah we don't need that commit, I can try this locally and push a change

danez:
I just tested this also locally and it doesn't work and stops with `EMFILE: too many open files`. Seems reading all the tests async without limit is not a good idea, but works on travis for some reason.

nicolo-ribaudo:
We could use [graceful-fs](https://github.com/isaacs/node-graceful-fs), which handles EMFILE errors automatically.

hzoo:
Yeah graceful-fs sounds good

hzoo:
Thanks for starting this @jugglinmike!

Makefile
```diff
@@ -30,3 +31,14 @@ bootstrap-flow: clean
 
 test-flow:
 	node scripts/run_flow_tests.js
+
+bootstrap-test262: clean
+	mkdir ./build
+	git clone https://github.com/tc39/test262.git ./build/test262
```
hzoo: could we do a depth=x here as well?

hzoo:
screen shot 2017-08-06 at 9 25 41 pm

jugglinmike:
It's been my pleasure :)

PR #658, opened 2017-08-03
Syntax Error: add message with the plugin that should be enabled

### LINKED ISSUE #169: Better error message on `unexpected token` when it's because a plugin isn't turned on

hzoo:
Fixes #169
Closes #178
Rebased/rewrite of https://github.com/babel/babylon/pull/178

Add a new method: `this.expectPlugin` instead of `this.hasPlugin` which throws if the plugin isn't enabled.

screen shot 2017-08-03 at 5 00 49 pm

This also means something better than "unexpected token" when it's a config issue. This also isn't tied to Babel but just babylon's plugins, so other tools can catch the error like Babel will to produce their own specific error messages.

screen shot 2017-08-03 at 5 01 51 pm

- [ ] non-block: make sure this covers all the relevant uses of `hasPlugin` (can be in later PRs)
  - [ ] bigInt, classPrivateProperties, exportExtensions, numericSeparator
- [ ] document the need to write a test without a plugin to get test coverage for each plugin added in the future


src/parser/statement.js
```diff
@@ -491,12 +486,13 @@ export default class StatementParser extends ExpressionParser {
     if (this.match(tt._catch)) {
       const clause = this.startNode();
       this.next();
-      if (this.match(tt.parenL) || !this.hasPlugin("optionalCatchBinding")) {
+      if (this.match(tt.parenL)) {
         this.expect(tt.parenL);
```
Kovensky: I guess this became redundant now?

src/parser/expression.js
```diff
@@ -627,11 +622,11 @@ export default class ExpressionParser extends LValParser {
         return this.finishNode(node, "Super");
 
       case tt._import:
-        if (this.hasPlugin("importMeta") && this.lookahead().type === tt.dot) {
+        if (this.lookahead().type === tt.dot) {
           return this.parseImportMetaProperty();
```
bakkot: Did you drop a `this.expectPlugin("importMeta")` here?

nicolo-ribaudo: It's called by the `parseImportMetaProperty` function.

hzoo: Yeah this problem was brought up by @motiz88, @danez in https://github.com/babel/babylon/pull/178/files#r87556780

If we make the error here I guess it's a different error if they just have `import.` vs `import.asdf` but maybe we shouldn't make a big deal out of it?

src/parser/expression.js
```diff
@@ -679,7 +674,9 @@ export default class ExpressionParser extends LValParser {
         return id;
 
       case tt._do:
-        if (this.hasPlugin("doExpressions")) {
+        // TODO
+        if (true) {
```
danez: What is this ? :D

src/parser/expression.js
```diff
@@ -808,6 +801,16 @@ export default class ExpressionParser extends LValParser {
     propertyName: string,
   ): N.MetaProperty {
     node.meta = meta;
+
+    if (meta.name === "function" && propertyName === "sent") {
+      if (this.isContextual(propertyName)) {
+        this.expectPlugin("functionSent");
+      } else if (!this.hasPlugin("functionSent")) {
+        // They didn't actually say `function.sent`, just `function.`, so a simple error would be less confusing.
+        this.unexpected();
```
danez: Uncovered. I add a test.

src/parser/expression.js
hzoo: there's a bug somewhere if I don't have it if you want to look into that 😄

Kovensky: Is this `if` necessary at all? The `_do` context can only be legally entered if the plugin is enabled to begin with.

danez:
I enabled the no-case-declarations. This should prevent this problems.

danez:
Okay I removed the if (true) and also rebased the branch.

src/parser/expression.js
hzoo: no it's not yeah I just need a block if that works

src/parser/statement.js
```diff
@@ -1145,17 +1145,14 @@ export default class StatementParser extends ExpressionParser {
...
+    if (!node.typeAnnotation) {
+      this.expectOnePlugin(["classProperties"]);
     }
 
     this.state.inClassProperty = true;
 
     if (this.match(tt.eq)) {
-      if (!this.hasPlugin("classProperties"))
-        this.raise(this.state.start, noPluginMsg);
+      this.expectOnePlugin(["classProperties"]);
```
nicolo-ribaudo: Here you can use `exceptPlugin`

src/parser/statement.js
```diff
@@ -1145,17 +1145,14 @@ export default class StatementParser extends ExpressionParser {
   }
 
   parseClassProperty(node: N.ClassProperty): N.ClassProperty {
-    const noPluginMsg =
-      "You can only use Class Properties when the 'classProperties' plugin is enabled.";
-    if (!node.typeAnnotation && !this.hasPlugin("classProperties")) {
-      this.raise(node.start, noPluginMsg);
+    if (!node.typeAnnotation) {
+      this.expectOnePlugin(["classProperties"]);
```
nicolo-ribaudo: Here you can use `exceptPlugin`

test/fixtures/experimental/_no-plugin/class-properties-with-initializer/options.json
```diff
@@ -0,0 +1,4 @@
+{
+  "throws": "This experimental syntax requires enabling one of the following parser plugin(s): 'classProperties, typescript' (2:6)",
```
nicolo-ribaudo: Is this influenced by #666?

PR #664, opened 2017-08-08
Fix conflict between "typescript" and "exportExtensions" plugins

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | #663
| License           | MIT

There was already a fix for this bug in `flow.js` by overriding `isExportDefaultSpecifier`, so did something similar in `typescript.js`.
Also needed to upgrade the `yarn.lock` due to getting an issue similar to babel/babel#5797: `path.arrowFunctionToExpression is not a function` in a clean checkout.


src/plugins/typescript.js
```diff
@@ -1186,6 +1186,19 @@ export default (superClass: Class<Parser>): Class<Parser> =>
     // OVERRIDES
     // ======================================================
 
+    isExportDefaultSpecifier(): boolean {
+      if (
+        this.match(tt.name) &&
+        (this.state.value === "type" ||
+          this.state.value === "interface" ||
+          this.state.value == "enum")
```
danez: ===

PR #666, opened 2017-08-08
Don't enable class properties just because "typescript" plugin is enabled

andy-ms:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | yes (breaking compared to another prerelease version)
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     |
| License           | MIT

As mentioned in https://github.com/babel/babel/pull/5899#discussion_r131689622, it seems more consistent to require the "classProperties" plugin, just like we require one for object rest/spread.


andy-ms:
Waiting on babel/babel#6074 for tests to pass

Issue #667, opened 2017-08-08
Disallow nested destructuring assignment with rest properties


gsathya:
### Input Code

```js
 ( {...{}} = {} ) 
```

### Babylon/Babel Configuration

stage3 preset

### Expected Behavior

SyntaxError: `...` must be followed by an assignable reference in assignment contexts

### Current Behavior

No error 

### Your Environment
[repl](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=stage-3&targets=Node-4&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&code_lz=BQbwdBIL5QBAvLaBKIA&playground=true)

nicolo-ribaudo:
Isn't this #661?

gsathya:
> Isn't this #661?

No, that's for binding pattern. This is for assignment pattern.

Andarist:
Is this still for grabs? I would like to work on it, but aint sure if noone has started yet.

nicolo-ribaudo:
Go for it 👍
You might also want to look for a fix for #661

Andarist:
Yeah, if I do this one I'll definitely look into other one, just don't want to block the other one yet, cause Im just diving into babylon for the first time :)

nicolo-ribaudo:
If you need any help, you can join our slack: https://slack.babeljs.io/

Andarist:
Can be closed, PR already merged in.

Issue #661, opened 2017-08-06
Disallow binding patterns of arrays or objects in BindingRestProperty


gsathya:
### Input Code

```js
let {...{}} = {};
```

### Babylon/Babel Configuration

stage3-preset

### Expected Behavior
SyntaxError: `...` must be followed by an identifier in declaration contexts

### Current Behavior
No error

### Your Environment

[repl](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=stage-3&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&code_lz=DYUwLgBA3gdHUF8EQLzQQbiA&playground=true)

danez:
This was already fixed and the latest versions of 6.x and 7.x both work for me. Not sure what the repl is using though.

gsathya:
> Also the repl seems to work for me successfully.

Does the link I pasted work? It doesn't throw a syntax error for me.

danez:
Ohh I mixed something up, I thought you expected it to pass and got an error, but you said the opposite. :)

hzoo:
Yeah it was already an error in 6.x and I remember at the meeting we decided to make it actually an error but it got "fixed" lol so need to revert in a sense

Andarist:
Can be closed, PR already merged in :)

hzoo:
Closed by https://github.com/babel/babylon/pull/670, thanks @Andarist!

PR #670, opened 2017-08-13
Adjusted Object Rest/Spread syntax handling to the latest spec

Andarist:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | #667, #661 
| License           | MIT

It's my first PR to babylon and in the current shape its rather really naive, although working for now (tests passing). Im going to comment specific lines with my comments/concerns - help would be appreciated :)


src/parser/expression.js
```diff
@@ -1118,8 +1118,11 @@ export default class ExpressionParser extends LValParser {
 
       if (this.hasPlugin("objectRestSpread") && this.match(tt.ellipsis)) {
         prop = this.parseSpread(isPattern ? { start: 0 } : undefined);
-        prop.type = isPattern ? "RestElement" : "SpreadElement";
-        if (isPattern) this.toAssignable(prop.argument, true, "object pattern");
+        if (isPattern) {
+          this.checkToRestConversion(prop);
```
Andarist: this probably could be forbidden somehow [here](https://github.com/babel/babylon/blob/88cf1bc140d19da19e9496b6293103d992bd9e85/src/parser/lval.js#L58) and [here](https://github.com/babel/babylon/blob/88cf1bc140d19da19e9496b6293103d992bd9e85/src/parser/lval.js#L53), but I would need to have knowledge there is im in declaration context **and** im checking rest's argument and Im not entirely sure whats the best practice to share such information

src/parser/lval.js
```diff
@@ -84,6 +84,8 @@ export default class LValParser extends NodeUtils {
           break;
 
         case "SpreadElement":
+          this.checkToRestConversion(node);
```
Andarist: this is supposed to fix https://github.com/babel/babylon/issues/667 , it **probably** could be checked inside of the `toAssignable` call 3 lines below, but as mentioned in the other comment - im not sure whats the best way to share parent's (in this example parent being rest element) context/state information

src/parser/lval.js
```diff
@@ -378,4 +380,18 @@ export default class LValParser extends NodeUtils {
       }
     }
   }
+
+  checkToRestConversion(node: SpreadElement): void {
+    const validArgumentTypes = [
```
Andarist: this should also be specified imho by the [RestElement's type](https://github.com/babel/babylon/blob/88cf1bc140d19da19e9496b6293103d992bd9e85/src/types.js#L594) but when I was trying to change it flow has yelled at me so much, that i just fled with my tails between my legs (for the time being - would love to adjust it properly)

src/parser/lval.js
nicolo-ribaudo: It's ok to check it here.

src/parser/lval.js
nicolo-ribaudo: It's ok to leave the type definition as is, because `RestElement` is also used by array destructuring (and the argument of a rest element inside an array pattern can be a pattern).

src/parser/lval.js
```diff
@@ -378,4 +380,18 @@ export default class LValParser extends NodeUtils {
...
+      "Identifier",
+      "MemberExpression",
+      "AssignmentPattern",
```
nicolo-ribaudo: I just checked and it looks this was already a bug in babylon... so can you add a test? 😊

src/parser/lval.js
Andarist: out of curiosity, does having rest's argument being an array pattern inside an array pattern have any purpose? besides that its possible to do ;p

it seems those are equivalents:
```js
// 1.
var [a, ...[b, ...c]] = [1,2,4,5,6]
```
```js
// 2.
var [a, b, ...c] = [1,2,4,5,6]
```
It seems to me that this is the same case which got now forbidden for object's rest:
```js
 ( {...{a, b}} = {} ) 
```
which simply didnt do much (to my understanding), just unwrapping/flattening this nested pattern

src/parser/lval.js
Andarist: Oh, yeah, Ive seen this as allowed so Ive added it as valid. Gonna remove this and add a test

src/parser/lval.js
```diff
@@ -378,4 +380,18 @@ export default class LValParser extends NodeUtils {
...
+    ];
+
+    if (validArgumentTypes.indexOf(node.argument.type) !== -1) {
```
Andarist: imho `validArgumentTypes` is more readable as its describing in the name why we check those and this dont have to be deduced by the reader, but I can refactor it to `&&` if u want

src/parser/lval.js
nicolo-ribaudo: `...[]` can be useful, for example, if you need to exaust an iterator:

```js
function* logAndYieldNumbers(n) {
  for (let i = 0; i < n; i++) { console.log(i); yield i; }
}

// Log all numbers, but keep only the first two.
let [ a, b, ...[] ] = logAndYieldNumbers(4);
```

src/parser/lval.js
nicolo-ribaudo: `validArgumentTypes` is ok then 👍

src/parser/lval.js
Andarist: ok... thats a weirdo case :trollface: but a valid one, thanks for sharing!

Andarist:
@nicolo-ribaudo ive pushed out everything that had to be pushed in this repo, so you can check it out

still need to change babel's test though, but my `make` went totally mad (`make: error: unable to find utility "make", not a developer tool or in PATH` 😂 ) so Ill tackle this in following days

existentialism:
@Andarist mind rebasing and updating the test262 whitelist again? I believe all the items you marked are fixed by https://github.com/tc39/test262/pull/1187

Andarist:
@existentialism I've rebases and reverted the whitelist additions, but there are still 2 problems:

- https://github.com/babel/babel/pull/6102 is not merged in, so integration test is failing

- we need to update `TEST262_COMMIT` so we test against latest test suite, however it gives me horrific results now 😨 

```
Summary:
 ✔ 51119 valid programs parsed without error
 ✔ 3043 invalid programs produced a parsing error
 ✔ 765 invalid programs did not produce a parsing error (and allowed by the whitelist file)
 ✔ 272 valid programs produced a parsing error (and allowed by the whitelist file)

 ✘ 484 valid programs produced a parsing error (without a corresponding entry in the whitelist file)
 ✘ 176 non-existent programs specified in the whitelist file
```

I guess there were many additions between previous commit we were pointing at and the latest master - https://github.com/tc39/test262/compare/4ea2931f169d4a4b5a9a4a7c731cc92bf7b3e13c...master

Have no idea what should be done now. Any recommendations?

jridgewell:
- The babel pr is merged now.
- That's fine, let's update the whitelist.

Andarist:
@jridgewell 

> That's fine, let's update the whitelist.

should I add 484 and remove 176 items to/from the whitelist?

jridgewell:
Yup. Maybe another PR? Just to keep this one clean.

Andarist:
@jridgewell 
https://github.com/babel/babylon/pull/692

please mention me when it gets merged, so I can rebase this one

existentialism:
@Andarist merged!

Andarist:
Rebased, checks passing, all good here! :)

existentialism:
@Andarist thanks!

PR #693, opened 2017-08-25
Raise error if trailing comma after rest element in async func params

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | y
| Breaking change?  | n
| New feature?      | n
| Deprecations?     | n
| Spec compliancy?  | y
| Tests added/pass? | y/y
| License           | MIT


danez:
I think trailing commas shouldn't be allowed anywhere after a RestElement, if I'm not mistaken.

src/parser/expression.js
```diff
@@ -494,14 +494,29 @@ export default class ExpressionParser extends LValParser {
...
         tt.parenR,
         possibleAsync,
+        refTrailingCommaPos,
       );
       this.finishCallExpression(node);
 
       if (possibleAsync && this.shouldParseAsyncArrow()) {
         state.stop = true;
+
+        if (refTrailingCommaPos.start > -1) {
```
jridgewell: Why do async and non-async parse their parameters differently?

nicolo-ribaudo: I believe they are parsed differently because non-async arrow function are parsed by converting a parenthesized sequence expression, while async functions by converting a call expression (e.g. `(arrow, or, parenthesized)` and `async(call, to, the = "async", function, or, async, arrow, parameters)`).

PR #696, opened 2017-08-26
Throw if rest element is not in last position for object expressions

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | y
| Breaking change?  | n
| New feature?      | n
| Deprecations?     | n
| Spec compliancy?  | y
| Tests added/pass? | y/y
| License           | MIT


src/parser/lval.js
```diff
@@ -124,6 +110,28 @@ export default class LValParser extends NodeUtils {
     return node;
   }
 
+  parseAssignableObjectExpressionProp(
```
nicolo-ribaudo: Can this be renamed to `toAssignableObjectExpressionProp`, since it checks and converts an _already parsed_ property?

PR #698, opened 2017-08-26
Rename PrivateName.name to .id

Qantas94Heavy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | not for release
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | N/A
| License           | MIT

Given that it refers to an identifier, it's somewhat confusing to have
to use `.name.name` to get the actual string. The type declared within
types.js is also incorrect, which this commit fixes.

This will also require anything using PrivateName in the main Babel repository to be changed, but it's probably better we do this now rather than after release.

### DIFF
--- a/ast/spec.md
+++ b/ast/spec.md
@@ -174,7 +174,7 @@ An identifier. Note that an identifier may be an expression or a destructuring p
 ```js
 interface PrivateName <: Expression, Pattern {
   type: "PrivateName";
-  name: Identifier;
+  id: Identifier;
 }
 ```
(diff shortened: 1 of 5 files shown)


hzoo:
Need to change https://github.com/babel/babylon/blob/master/ast/spec.md#privatename too then

Qantas94Heavy:
Done.

PR #699, opened 2017-08-26
Breaking: Apply location data fix to decorators plugin

JamesHenry:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 694
| License           | MIT

This "backports" the location data fix from the `decorators2` plugin to the original `decorators` plugin.


jseminck:
This change indeed correctly modifies the `ClassExpression` range. But if you use `export default` on that class, then the `ExportDefaultDeclaration` range is not updated.

This is with the latest version of babylon7:

screen shot 2017-08-31 at 22 23 15

Is this behaviour correct/expected? This currently causes an issue on ASTExplorer.net - but I'd first like to understand if the AST is correct before thinking about a possible change to the ASTExplorer highlighting algorithm: https://github.com/fkling/astexplorer/issues/269

As always, thanks for the support/work you guys do for babel 🎉

hzoo:
original decorators pr is https://github.com/babel/babylon/pull/587

JamesHenry:
@jseminck Can you please confirm the behaviour with the original PR? AFAICT all I did was backport the fix, so it would suggest that the fix was missing this case

danez:
This breaks comment attachment. Comments that have been previously been attached to the decorator are now attached to the class

```js
/**
 * Decorator description
 */
 @Decorator1
class A {}
```

jseminck:
@JamesHenry sorry for the slow reply. I'm quite sure the behaviour is the same with the original change... I'm trying to dig into the code to figure out how it could potentially be changed (I'm really not certain if it's even a bug yet).

PR #703, opened 2017-08-29
Add support for class private methods

Qantas94Heavy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

This adds support for private methods, as described in https://github.com/littledan/proposal-private-methods.

I haven't made an issue for this proposal yet, but if we want it to be separate from #540, I can do so.

  - [x] An update to the [#plugins](https://github.com/babel/babylon#plugins) part of the readme. Add a new entry to that list for the new plugin flag (and link to the proposal)
  - [x] If any new nodes or modifications need to be added to the AST, update [ast/spec.md](https://github.com/babel/babylon/blob/master/ast/spec.md)
  - [x] Make sure you use the `this.hasPlugin("plugin-name-here")` check so that your new plugin code only runs when that flag is turned on (not default behavior)
  - [x] Add failing/passing tests according to spec behavior

I'll also need to submit changes to `babel-types` to match up with this PR.

**NOTE:** There's still a little bit I'm not exactly sure about (tick means handled):

 - [x] There's some points where this code "eats" the `#` then throws the error, while in other parts it throws the error without "eating" the `#`.
 - [x] This might need a few more tests.
 - [x] One of the existing ASI tests seems to fail (fixtures/experimental/class-private-properties/asi-failure-inline) because the error message doesn't include `expected ;`, but the error is still thrown in the same position in the test file. **(edit: changed test to match new message)**


src/parser/statement.js
```diff
@@ -1009,12 +1019,36 @@ export default class StatementParser extends ExpressionParser {
...
 
     this.parsePostMemberNameModifiers(methodOrProp);
 
     if (this.isClassMethod()) {
+      if (isPrivate) {
+        // TODO: seems a bit inconsistent error throwing: in some cases we've already
+        // eaten the #, but in other cases we haven't. How to solve?
+        this.expectPlugin("classPrivateMethods");
+        // private "normal" method
+        method.kind = "method";
```
jridgewell: Can we hoist this to the beginning of the `this.isClassMethod()` if statement?

src/parser/statement.js
```diff
@@ -1009,12 +1019,36 @@ export default class StatementParser extends ExpressionParser {
       return;
     }
 
+    const isPrivate = this.eat(tt.hash);
+
+    if (isPrivate) {
+      this.expectOnePlugin(["classPrivateProperties", "classPrivateMethods"]);
```
jridgewell: I think we need to do this before eating the hash.

src/parser/statement.js
```diff
@@ -1069,27 +1129,52 @@ export default class StatementParser extends ExpressionParser {
       // `get\n*` is an uninitialized property named 'get' followed by a generator.
       // a getter or setter
       method.kind = key.name;
-      this.parsePropertyName(method);
-      if (this.isNonstaticConstructor(method)) {
-        this.raise(method.key.start, "Constructor can't have get/set modifier");
+      if (this.match(tt.hash)) {
+        // private getter/setter
+        this.expectPlugin("classPrivateMethods");
+        // The private # wouldn't have been eaten as the "async" was in front of it.
```
jridgewell: Comment is incorrect.

src/parser/statement.js
```diff
@@ -1069,27 +1129,52 @@ export default class StatementParser extends ExpressionParser {
...
+        this.next();
+        // The so-called parsed name would have been "async": get the real name.
```
jridgewell: Comment again.

src/parser/statement.js
```diff
@@ -1069,27 +1129,52 @@ export default class StatementParser extends ExpressionParser {
...
+          );
+        }
+        this.parseClassMethod(
+          classBody,
+          method,
+          false,
+          false,
+          /* isConstructor */ false,
+        );
+        this.checkGetterSetterParamCount(method);
```
jridgewell: I think this should apply to private accessors, too.

src/types.js
```diff
@@ -664,6 +666,13 @@ export type ClassMethod = MethodBase &
     variance?: ?FlowVariance, // TODO: Not in spec
   };
 
+export type ClassPrivateMethod = NodeBase &
+  ClassMethodOrDeclareMethodCommon & {
+    type: "ClassPrivateMethod",
+    key: Identifier,
+    static: false,
```
jridgewell: Should be `boolean`, right?

src/parser/statement.js
```diff
@@ -1134,17 +1219,14 @@ export default class StatementParser extends ExpressionParser {
...
   ): N.ClassPrivateProperty {
     this.state.inClassProperty = true;
 
-    if (this.match(tt.eq)) {
-      this.next();
-      node.value = this.parseMaybeAssign();
-    } else {
-      node.value = null;
-    }
+    if (node.computed !== undefined) node.computed = undefined;
```
jridgewell: What's this for?

ast/spec.md
```diff
@@ -1049,6 +1050,19 @@ interface ClassMethod <: Function {
...
+## ClassMethod
+
+```js
+interface ClassPrivateMethod <: Function {
+  type: "ClassPrivateMethod";
+  key: Identifier;
+  kind: "method" | "get" | "set";
+  computed: boolean;
+  static: boolean;
+  decorators: [ Decorator ];
```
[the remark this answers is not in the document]
Qantas94Heavy: Would that be needed for the normal `ClassMethod` as well?

src/parser/statement.js
Qantas94Heavy: Confused about where exactly you want this part to be moved to.

src/parser/statement.js
Qantas94Heavy: Huh? Maybe my wording was confusing.

What I'm trying to say is that it's now being eaten, but it wasn't detected as being "private" in the code above (`let isPrivate = ...`). Is that wrong?

src/types.js
Qantas94Heavy: Fixed.

ast/spec.md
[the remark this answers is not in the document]
jridgewell: Yup.

src/parser/statement.js
jridgewell: Before the `if (isPrivate)`.

src/parser/statement.js
jridgewell: The "async" part is incorrect. We're in the "get"/"set" path here.

src/types.js
jridgewell: Oh, it's already in `ClassMethodOrDeclareMethodCommon`, so we can just drop this line.

src/parser/statement.js
Qantas94Heavy: Done.

src/parser/statement.js
```diff
@@ -1009,21 +1019,43 @@ export default class StatementParser extends ExpressionParser {
...
+      : this.parseClassPropertyName(methodOrProp);
 
     this.parsePostMemberNameModifiers(methodOrProp);
 
     if (this.isClassMethod()) {
+      method.kind = "method";
+
+      if (isPrivate) {
+        // TODO: seems a bit inconsistent error throwing: in some cases we've already
+        // eaten the #, but in other cases we haven't. How to solve?
```
nicolo-ribaudo: You could make `expectPlugin` accept an optional `pos` parameter (like `unexpected` does) and pass `node.start` to it.

src/parser/statement.js
```diff
@@ -1009,21 +1019,43 @@ export default class StatementParser extends ExpressionParser {
       return;
     }
 
+    const isPrivate = this.match(tt.hash);
+    if (isPrivate) {
+      this.expectOnePlugin(["classPrivateProperties", "classPrivateMethods"]);
```
nicolo-ribaudo: Instead of throwing here, you can wait until it is known if the node is a property or a method (to give a more specific error message).

src/parser/statement.js
```diff
@@ -1188,6 +1270,24 @@ export default class StatementParser extends ExpressionParser {
     );
   }
 
+  parseClassPrivateMethod(
+    classBody: N.ClassBody,
+    method: N.ClassMethod,
+    isGenerator: boolean,
+    isAsync: boolean,
+    isConstructor: boolean,
```
nicolo-ribaudo: Can this ever be `true`?

bakkot: No. Not only that, it should be a syntax error; see below.

src/parser/statement.js
Qantas94Heavy: Yeah, that was a dodgy hack for some test failure because I had set `computed` elsewhere by accident. I've fixed this now.

src/parser/expression.js
```diff
@@ -1398,20 +1398,26 @@ export default class ExpressionParser extends LValParser {
...
       prop.key =
         this.match(tt.num) || this.match(tt.string)
           ? this.parseExprAtom()
-          : this.parseIdentifier(true);
+          : this.parseMaybePrivateName();
+
+      // TODO: if no plugin and is PrivateName, should error be handled here or elsewhere?
+      if (prop.key.type !== "PrivateName") {
+        // ClassPrivateProperty is never computed, so we don't assign in that case.
+        prop.computed = false;
```
jridgewell: `prop` here is a `MemberExpression`? In that case, it should always be set to `false`.

src/parser/statement.js
```diff
@@ -976,54 +965,86 @@ export default class StatementParser extends ExpressionParser {
...
-    const isSimple = this.match(tt.name);
     const key = this.parseClassPropertyName(methodOrProp);
+    let isPrivate = key.type === "PrivateName";
+    const isSimple = !isPrivate && !key.computed;
+
+    const handleClassProperty = () => {
+      if (isPrivate) {
+        this.expectPlugin("classPrivateProperties", key.start);
+        // Private property
+        classBody.body.push(this.parseClassPrivateProperty(prop));
```
jridgewell: Can we make a `#pushClassPrivateProperty`?

src/parser/statement.js
```diff
@@ -976,54 +965,86 @@ export default class StatementParser extends ExpressionParser {
...
+      method.kind = "method";
+
+      if (isPrivate) {
+        this.expectPlugin("classPrivateMethods", key.start);
+        this.parseClassPrivateMethod(
+          classBody,
+          method,
+          false,
+          false,
+          isConstructor,
```
jridgewell: Can't be `true`.

src/parser/expression.js
Qantas94Heavy: I think the method signature says it's a `Class[Private]Property` or `Class[Private]Method`.

src/parser/statement.js
Qantas94Heavy: Done -- should the two be merged or remain separate?

Qantas94Heavy:
I'm just wondering whether the one test failure is important or not.

This roughly equivalent non-private code just throws `Unexpected token (2:4)`:

```
class A {
  a b;
}
```

Do we need it to check whether a semicolon should be there, and if so, how would I go about doing that?

bakkot:
@Qantas94Heavy, I don't think it's important that the failure says `Unexpected token, expected ;` vs just `Unexpected token`, especially given that the latter is what happens with regular properties. I'd just fix the test.

src/parser/statement.js
Qantas94Heavy: Removed (not sure why it's not saying it's outdated).

src/parser/statement.js
```diff
@@ -976,54 +965,72 @@ export default class StatementParser extends ExpressionParser {
...
     }
 
-    const isSimple = this.match(tt.name);
     const key = this.parseClassPropertyName(methodOrProp);
+    const isPrivate = key.type === "PrivateName";
+    // Check the key is not a computed expression or string literal.
+    const isSimple = key.type === "Identifier";
+
+    // TODO: should these be merged into pushClassProperty?
+    const handleClassProperty = () => {
```
jridgewell: This could just be inlined into the callsites.

Qantas94Heavy:
@jridgewell right, updated.

PR #712, opened 2017-09-03
Proposal: Revert AST for directives to conform with ESTree again

### LINKED ISSUE #216: String literal parsed incorrectly when containing "\\"

### LINKED ISSUE #289: Incorrect parsing of escaped DirectiveLiteral string

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes (In the long run)
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | Fixes #616, Fixes #289, 
| License           | MIT

As this is breaking it needs changes to babel and minify in order to work. Will do this if we agree on this.

Related bugs that will be fixed when this is applied and introduced to the rest of the babel stack: babel/babel#4767, babel/babili#308, babel/babel#5935.

# Problem

Currently we use the Nodes `Directive` and `DirectiveLiteral` to represent directives in the AST. This nodes are added to a `directives` field in `Program` or `BlockStatements`.

`DirectiveLiteral` looks very similar if not exactly the same as `StringLiteral`, but there are major differences:
  * The `value` of a StringLiteral is the parsed utf-8 representation of the original value
  * The `value` of a DirectiveLiteral is the **raw** value with quotes stripped.

Because the look so similar most babylon consumers (including babel) treat `DirectiveLiteral` simply as `StringLiteral`. This works in 95% of the cases but as soon as there are escape sequences involved the result will not be the same as can be seen in the linked issues.

For example:

```js
"use\\x20strict";
```

If it would be parsed as StringLiteral it would look like this:

```json
{
  "type": "StringLiteral",
  "value": "use strict"
}
```

In comparison the `DirectiveLiteral` would look like:

```json
{
  "type": "DirectiveLiteral",
  "value": "use\\\\x20strict"
}
```

# Proposed Solution

Instead of making sure every consumer of babylons AST is handling directives correctly I propose to drop the current handling of directives (introduced in v6.0.0) and revert to the ESTree specification in this case.

The ESTree specification [describes](https://github.com/estree/estree/pull/152) a field `directive` on `ExpressionStatement`s which contains the raw directive value. Other than that nothing changes, which is one of the good things, as this ensures that directives can be treated the same way as `StringLiterals` because they are `StringLiteral` nodes.

So looking at above example the current output is

```json
{
  "type": "Program",
  "body": [],
  "directives": [
    {
      "type": "Directive",
      "value": {
        "type": "DirectiveLiteral",
        "value": "use\\\\x20strict",
        "extra": {
          "raw": "\"use\\\\x20strict\"",
        }
      }
    }
  ]
}
```

The output with this PR is:

```json
{
  "type": "Program",
  "body": [
    {
      "type": "ExpressionStatement",
      "expression": {
        "type": "StringLiteral",
        "value": "use strict",
        "extra": {
          "raw": "\"use\\\\x20strict\"",
        }
      },
      "directive": "use\\\\x20strict"
    }
  ]
}
```

With the new AST it is not much harder tocheck if certain directives are present and can simply done by doing:

```js
function isStrictBody(node: Function|Program) {
  for (const statement of node.body.body) {
    if (statement.directive == null) {
     break;
    } else if (statement.directive === "use strict") {
      return true;
    }
  }

  return false;
}
```

In a lot of cases the babel code will also get cleaner, for example in cases where we right now have to always make sure that directives are carried correctly when a body is moved. This will not have to be done anymore, as directives are part of the body.

As this is a major breaking change it would be nice to get some input.

Maybe also from @kittens as you have created the current behaviour and might have some insight if it would be okay or if there are usecases that I'm missing.


bakkot:
For what it's worth, I disagree that "directives can be treated the same way as StringLiterals" is a good thing. Every time any project whose authors are not familiar with that particular part of the JS spec tries to do so they up with a bug where they treat `"use strict"` like `"use\x20strict"` and/or `("use strict")`. I don't think it's a good idea to encourage that by making them appear the same.

Anyway, given that https://github.com/estree/estree/pull/152 hasn't been merged yet, why not have ESTree conform to Babylon, instead of the other way around? Babylon's approach is better; there's still time to do the right thing in ESTree!

danez:
This change was never about getting closer to the ESTree spec, it just happened that the spec has a different way of handling directives that would solve a lot of the problems we are facing and I also never said any of the arguments mentioned. No need to get emotional. :)

@bakkot I agree that directives and string literals should be treated different when evaluating them, but right now directives are treated as string literals when printing although the shouldn't . Syntax wise they are string literals with some restrictions. And in the case we're they are parsed and later printed they should be able to be treated as String literals in my opinion, because otherwise you always end up having a print method that handles string literals and one for directives which you need to make sure always generate the same output for the same string represented differently in different nodes.
I see your argument makes sense, but in my opinion it is more important that people also print directives the correct way, which does not happen right now, because of the above reasons. It boils down to the same issue that if someone does not know the spec and which field in a node represents which value then you always end up with misinterpretions.

bakkot:
@danez I think the current setup is a lot more likely to lead to people doing the right thing than estree/estree#152 is. You *can't* treat a directive like a string literal when printing and the AST should not encourage you to do so; you *need* different methods to print them.

Moreover, there's no reason the AST should represent a.) the "cooked" value of a directive or b.) the "raw" value of a non-directive string literal. Putting `"use strict"` anywhere in the tree when parsing `(function(){"use\x20strict";})` is at best extremely misleading.

loganfsmyth:
I'd much rather keep the directive type and make it raw-only instead so it's clear that it isn't a string literal.

> As this is breaking it needs changes to babel and minify in order to work.

I'm not sure we've clarified just how big a task this would be. This means _anywhere_ that does `.body.unshift` or `.unshiftContainer('body'` needs to essentially be deleted and rewritten to take directives into account. That's a massive hit to the conceptual model in the API I think, since no-one expects to have to worry about directives right now.

> The ESTree specification describes a field directive on ExpressionStatements which contains the raw directive value.

This approach works for read-only usecases, but as soon as it hits transformation, the separation is essentially meaningless. Semantically a `Directive` node _must_ be at the start, or it isn't a directive anymore. ESTree's desire to be backward-compatible is forcing that representation, but it's not a good one for actually representing the semantics of the code over the source of transformation, since you could easily traverse for all Directive nodes only to find that they aren't actually directives anymore because someonething else got inserted before them. The AST should try to enforce this stuff as early as possible.

danez:
@loganfsmyth Good points I haven't thought about this cases where you insert at the top of the block. 👍 
What do you mean with raw-only? That's what it is right now, but having the raw value in the `value` field.

loganfsmyth:
Yeah I guess my suggestion is that we stop printing `DirectiveLiteral` as `StringLiteral`, instead doing

```
export function DirectiveLiteral(node) {
  const q = node.value.indexOf('"') ? '\'' : '"';
  this.token(q + node.value + q);
}
```

or something along those lines.

danez:
Okay I create a PR.

PR #731, opened 2017-09-23
Clean up uncatagorised decorators (1/2)

pudility:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | N/A
| Tests added/pass? | yes
| Fixed tickets     | None
| License           | MIT

I moved uncatagorised 34-40 into decorators. I did this to try to clean up the uncategorised directory a little bit.


test/fixtures/experimental/uncategorised/33/actual.js
```diff
@@ -1 +0,0 @@
-@foo class Foo {}
```
jridgewell: Where'd this test go?

pudility: It was almost the same as `decorators/class-decorator/actual.js`.  https://github.com/babel/babylon/blob/bfd9c7ba508cf06813636e2dc6adf5658c743b34/test/fixtures/experimental/decorators/class-decorator/actual.js#L1-L4

nicolo-ribaudo: I think we can keep both, since in `decorators/class-decorator/actual.js` the decorator is on a different line and thus have different locations.

pudility: Alright, I added it back.

nicolo-ribaudo: It is `same-line`

pudility: I am **so** glad that I have other people check my work. Sorry about that 😳.

pudility:
Just a quick question, if I make more organizing PR should they be bigger or smaller? For example, I am thinking of categorising the rest of uncategorisd/40-54. Should these be two PR or just one?

jridgewell:
This is a good size.

pudility:
Ok, Thanks.

PR #740, opened 2017-09-27
Add throw expressions

jridgewell:
Stage 2 proposal: https://github.com/tc39/proposal-throw-expressions.


nicolo-ribaudo:
"throw" should also be added here:
https://github.com/babel/babylon/blob/master/src/types.js#L419-L426

jridgewell:
Done.

src/types.js
```diff
@@ -423,7 +423,8 @@ export type UnaryOperator =
   | "~"
   | "typeof"
   | "void"
-  | "delete";
+  | "delete",
```
nicolo-ribaudo: I think the comma here is illegal

jridgewell: Whoops.

PR #741, opened 2017-09-28
Don't build for tags

hzoo:
sounds good? should do this everywhere

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,6 @@ after_success:
   - 'if [ "$JOB" = "test-coverage" ]; then bash <(curl -s https://codecov.io/bash) -f coverage/coverage-final.json -F babylon ; fi'
 
 notifications:
-  email: false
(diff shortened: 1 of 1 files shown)


danez:
Why?

hzoo:
Do you use the emails? slack pings build status already, and tag ci isn't necessary since it already builds to master

danez:
I was never pinged by slack before. For me emails is the only way currently to get notified if my PRs fail, but if I can somehow setup slack I'm happy with that.

.travis.yml
```diff
@@ -51,4 +51,9 @@ after_success:
   - 'if [ "$JOB" = "test-coverage" ]; then bash <(curl -s https://codecov.io/bash) -f coverage/coverage-final.json -F babylon ; fi'
 
 notifications:
+  email: false
```
jridgewell: 👎

hzoo:
K i'l remove the email thing

PR #742, opened 2017-09-28
Pipeline Operator proposal

hzoo:
Ref https://github.com/babel/proposals/issues/29
Taken from https://github.com/babel/babel/pull/3159 by @gilbert
Spec: https://github.com/tc39/proposal-pipeline-operator/pull/51/files

### need more tests
- [x] precedence
- [x] with arrow function `10 |> x => x + 1;`
- [x] chaining `a |> b |> c`
- [x] multline

```js
let result = "hello"
  |> doubleSay
  |> capitalize
  |> exclaim;
```

```js
let newScore = person.score
  |> double
  |> _ => add(7, _)
  |> _ => boundScore(0, 100, _);
```


src/tokenizer/index.js
```diff
@@ -411,9 +411,17 @@ export default class Tokenizer extends LocationParser {
     const next = this.input.charCodeAt(this.state.pos + 1);
     if (next === code)
       return this.finishOp(code === 124 ? tt.logicalOR : tt.logicalAND, 2);
+    if (code === 124) {
+      // '|>'
+      if (next === 62) {
+        this.expectPlugin("pipelineOperator");
```
jridgewell: Don't we usually expect in the not-tokenizer part?

src/tokenizer/index.js
```diff
@@ -411,9 +411,17 @@ export default class Tokenizer extends LocationParser {
     const next = this.input.charCodeAt(this.state.pos + 1);
```
jridgewell: Please rename the function.

test/fixtures/experimental/pipeline-operator/base/actual.js
```diff
@@ -0,0 +1 @@
+a |> b
```
jridgewell: More examples, especially arrows.

src/tokenizer/index.js
hzoo: this is ok now

test/fixtures/experimental/pipeline-operator/base/actual.js
hzoo: updated

src/tokenizer/types.js
```diff
@@ -61,7 +61,7 @@ export class TokenType {
     this.isAssign = !!conf.isAssign;
     this.prefix = !!conf.prefix;
     this.postfix = !!conf.postfix;
-    this.binop = conf.binop || null;
+    this.binop = conf.binop === 0 ? 0 : conf.binop || null;
```
littledan: This will be nice with ?? :)

src/parser/expression.js
```diff
@@ -290,6 +290,13 @@ export default class ExpressionParser extends LValParser {
 
         const startPos = this.state.start;
         const startLoc = this.state.startLoc;
+
+        if (node.operator === "|>") {
+          this.expectPlugin("pipelineOperator");
+          // Support syntax such as 10 |> x => x + 1
+          this.state.potentialArrowAt = startPos;
```
littledan: I don't think the current draft spec has this logic. I couldn't figure out how to put it into the grammar. Right now, `x |> y => z` would be a syntax error. But it's clearly useful--i'll think more about the spec.

hzoo: Good to know

src/tokenizer/types.js
hzoo: Yes, should be easy to add/use here (would be good to find this kind of code on github/codebase and then do a codemod)

PR #745, opened 2017-09-28
Fix: major revision to valid and invalid numeric literal separator "sibling" characters

rwaldron:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | yes
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| License           | MIT

As a result of testing the Babel transform against NLS tests pending for test262, I discovered a pretty serious oversight in my original parsing. This addresses that issue and adds a substantial amount of tests.


src/tokenizer/index.js
```diff
@@ -43,6 +43,29 @@ const forbiddenNumericSeparatorSiblings = {
...
+    48, 49, 50, 51, 52, 53, 54, 55, 56, 57,
+  ],
+  oct: [
+    // 0 - 7
+    48, 49, 50, 51, 52, 53, 54, 55,
+  ],
+  bin: [
+    // 0 - 1
+    48, 49,
+  ],
```
nicolo-ribaudo: Nit: I would like to avoid repeating all the numbers... What do you think about doing something like this?
```js
const allowedNumericSeparatorSiblings = {};
allowedNumericSeparatorSiblings.bin = [ 48, 49 ];
allowedNumericSeparatorSiblings.oct = [
  ...allowedNumericSeparatorSiblings.bin, 50, 51, 52, 53, 54, 55,
];
allowedNumericSeparatorSiblings.dec = [
  ...allowedNumericSeparatorSiblings.oct, 56, 57,
];
```

It wouldn't have a too big performance impact, since they would be computed only when babylon is loaded.
It's up to you, anyway.

rwaldron: Sure, I can do that.

rwaldron:
Looks like I have to update to use `indexOf`!

nicolo-ribaudo:
@rwaldron Can you fix the linting errors? `npm/yarn run lint -- --fix`

hzoo:
Looks nice the way it is, but prettier wants the newlines..

hzoo:
Thanks @rwaldron!!

rwaldron:
Sorry I missed these follow-up messages, I've been afk since Thursday evening.

Issue #747, opened 2017-09-30
Set id to null for ArrowFunctionExpressions?


EtherTyper:
Should the [ArrowFunctionExpression section of the AST specification](https://github.com/babel/babylon/blob/master/ast/spec.md#arrowfunctionexpression) clarify that supplying an `Identifier` for `id` does nothing, by explicitly saying `id: null`? Of course, we don't need to change `babel-generator` or `babel-types` to accommodate this change, because that would only hurt backwards compatibility.

Thanks,
Eli

hzoo:
Hey @EtherTyper! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

danez:
Thanks for opening this issue, would you mind also proposing this to https://github.com/estree/estree. We for the most part use their spec 1:1.

PR #748, opened 2017-10-01
Specify id to be null for arrow functions.

EtherTyper:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | no
| Fixed tickets     | #747
| License           | MIT

The [ArrowFunctionExpression section of the AST specification](https://github.com/babel/babylon/blob/master/ast/spec.md#arrowfunctionexpression) should probably clarify that supplying an `Identifier` for `id` does nothing, by explicitly saying `id: null`. Of course, we don't need to change `babel-generator` or `babel-types` to accommodate this change, because that would only hurt backwards compatibility.

Thanks,
Eli

### DIFF
--- a/ast/spec.md
+++ b/ast/spec.md
@@ -626,6 +626,7 @@ A `this` expression.
 ```js
 interface ArrowFunctionExpression <: Function, Expression {
   type: "ArrowFunctionExpression";
+  id: null;
   body: BlockStatement | Expression;
   expression: boolean;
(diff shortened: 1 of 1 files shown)


danez:

Thanks for your effort.

PR #755, opened 2017-10-11
Add JSX fragment syntax support

clemmy:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

This PR adds support for https://github.com/facebook/jsx/pull/93.

I'm also currently working on corresponding Babel PR as well, just leaving this up here first for review, since I'm not sure if I'm abusing `$FlowIgnore`. 😆 

UPDATE: Both PR's ready for review and Babel one can be found at https://github.com/babel/babel/pull/6552


src/plugins/jsx/index.js
```diff
@@ -37,6 +37,19 @@ tt.jsxTagEnd.updateContext = function(prevType) {
   }
 };
 
+function isFragment(
+  object:
+    | ?N.JSXOpeningElement
+    | ?N.JSXClosingElement
+    | ?N.JSXOpeningFragment
+    | ?N.JSXClosingFragment,
```
Kovensky: Fragments cannot be self-closing, but it might be good to just accept any `N.JSXElement` to make it easier to test against arbitrary elements.

Although this is an internal use only function, so maybe it doesn't need to be extra ergonomic 🤔

src/plugins/jsx/index.js
```diff
@@ -37,6 +37,19 @@ tt.jsxTagEnd.updateContext = function(prevType) {
...
+): boolean {
```
Kovensky: Does Flow support type narrowing similarly to TypeScript? If it does, the return type could be something similar to `: object is N.JSXOpeningFragment | N.JSXClosingFragment`.

src/plugins/jsx/index.js
```diff
@@ -410,8 +431,18 @@ export default (superClass: Class<Parser>): Class<Parser> =>
 
         if (
           // $FlowIgnore
```
Kovensky: All these `$FlowIgnore` are scary...... but they were already in place 🤷‍♀️

src/types.js
```diff
@@ -799,6 +799,9 @@ export type JSXAttribute = Node;
 export type JSXOpeningElement = Node;
 export type JSXClosingElement = Node;
 export type JSXElement = Node;
+export type JSXOpeningFragment = Node;
+export type JSXClosingFragment = Node;
+export type JSXFragment = Node;
```
Kovensky: Ah, that's why there are so many `$FlowIgnore` in the JSX parser. None of these types are written 😆

(Even if you feel inclined to, don't do it on this PR, of course)

test/fixtures/jsx/basic/fragment-2/actual.js
```diff
@@ -0,0 +1 @@
+<>Hi, I'm a string!</>
```
Kovensky: This will be very confusing for syntax highlighters for a while...

clemmy:
@Kovensky Hopefully https://github.com/facebook/jsx/pull/94 will be merged soon, so that my PR for Babel/Babylon can include those changes. :)

src/plugins/jsx/index.js
clemmy: Not that I can find in the docs! :(

clemmy:
Sorry I've been inactive with this PR for a while! I'm going to open a corresponding PR for this on the Babel repository in the coming days. :)

clemmy:
:man_dancing:

PR #757, opened 2017-10-12
Fix export extensions handling of TS declare keyword

### LINKED ISSUE #753: Typescript: ExportNamedDeclaration raises SyntaxError when exportExtensions plugin enabled
When parsing typescript code with the `exportExtensions` plugin enabled, code that is otherwise parseable fails to parse. My suspicion given the parsing logic behind export statements is due to the fact that this is because the `exportExtensions` plugin expects a specific token pattern, which is not the case when parsing an `ExportNamedDeclaration` statement.

### Input Code

```typescript
// input.ts
export declare interface Foo {
  a: number;
}
```
[ASTExplorer link](https://astexplorer.net/#/gist/b250659b4005044564f790cb20b45b7a/f786b95061028c40070b6f93853fe34854e51714)

```js
// Options given to babylon.parse
{
  sourceType: 'module',
  plugins: [
    'typescript',
    'classProperties',
    'objectRestSpread',
    'decorators',
    'exportExtensions',
    'asyncGenerators',
  ],
}
```

### Expected Behavior

The code should parse successfully, as this is valid typescript code, and the typescript plugin has been enabled. 

### Current Behavior

The code fails to parse, failing with a `SyntaxError: Unexpected Token` at the `loc` directly after the `declare` token. 

#### Additional Notes

Taking a look at the stack trace, it seems like the code that causes the parse failure occurs at [src/parser/statement.js:1296](https://github.com/babel/babylon/blob/664ff37b3651e189170d533d262455a74502a77b/src/parser/statement.js#L1296)

Note that this code is executed iff the `exportExtensions` plugin is enabled. My naive hunch is it seems like what the code is saying is:

> "First I expect an `export` token, followed by an export specifier, followed _immediately_ by a `from` token"

Perhaps it's getting confused when it encounters `declare` as it expects a symbolic identifier?

Or perhaps there may be a parsing issue within the [isDefaultExport() logic](https://github.com/babel/babylon/blob/664ff37b3651e189170d533d262455a74502a77b/src/parser/statement.js#L1350-L1352)(?)

Note that _the code parses successfully when `exportExtensions` is not enabled_.

### Your Environment

| software         | version
| ---------------- | -------
| Babylon          |  7.0.0-beta.27
| node             | v8.1.3
| npm              | 5.3.0
| Operating System | Linux (I unfortunately am not at liberty to give away any more information than this)

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | no
| Tests added/pass? | yes/yes
| Fixed tickets     | Fixes #753
| License           | MIT

Also noticed some inconsistencies with default exports unrelated to `exportExtensions` (will follow up with another PR).

### DIFF
--- a/src/plugins/typescript.js
+++ b/src/plugins/typescript.js
@@ -1191,20 +1191,29 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return this.finishNode(node, "TSTypeParameterInstantiation");
     }
 
+    tsIsDeclarationStart(): boolean {
+      if (this.match(tt.name)) {
+        switch (this.state.value) {
+          case "abstract":
+          case "declare":
+          case "enum":
+          case "interface":
+          case "module":
+          case "namespace":
+          case "type":
+            return true;
+        }
+      }
+
+      return false;
+    }
+
     // ======================================================
     // OVERRIDES
     // ======================================================
 
     isExportDefaultSpecifier(): boolean {
-      if (
-        this.match(tt.name) &&
-        (this.state.value === "type" ||
-          this.state.value === "interface" ||
-          this.state.value === "enum")
-      ) {
-        return false;
-      }
-
+      if (this.tsIsDeclarationStart()) return false;
       return super.isExportDefaultSpecifier();
     }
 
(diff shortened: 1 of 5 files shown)


src/plugins/typescript.js
```diff
@@ -1191,20 +1191,29 @@ export default (superClass: Class<Parser>): Class<Parser> =>
       return this.finishNode(node, "TSTypeParameterInstantiation");
     }
 
+    tsIsContextualExportable(): boolean {
```
existentialism: Any better name suggestions? 🙃

nicolo-ribaudo: `tsStartsDeclaration`? 🤔

PR #759, opened 2017-10-13
Allow new.target in class properties

### LINKED ISSUE #758: static field initializer as an arrow function containing new.target is not allowed
### Input Code

```js
class X {
  static c = () => new.target;
}
```

stage2, stage3

### Expected Behavior

No error

### Current Behavior

```
repl: new.target can only be used in functions (2:23)
  1 | class X {
> 2 |   static c = () => new.target;
    |                        ^
  3 | }
```
### Your Environment

[REPL](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=false&code_lz=MYGwhgzhAEAa0G8BQ1oQC5nQS2NPAvNABQCU0BAfNAHYCmA7gHSYBOA5negNxIC-QA&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=stage-2%2Cstage-3&targets=&version=7.0.0-beta.2)

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | Fixes #758 
| License           | MIT


nicolo-ribaudo:
Is `new.target` allowed in arrow parameters? e.g. `class A { x = (p = new.target) => {} }`

Also, I think `class A { x = (new.target) }` should throw.

existentialism:
Related: https://github.com/babel/babel/issues/6243

littledan:
`new.target` would always be `undefined`, just like in a method.

littledan:
Would it make sense to add a test that verifies that, when you execute the code, you actually get undefined for new.target in these cases?

nicolo-ribaudo:
I think we should add them, but they must be in the Babel repository (after this PR is merged)

nicolo-ribaudo:
LGTM

hzoo:
This will be better when babylon is merged into the monorepo

PR #761, opened 2017-10-14
Implement nullish coalescing operator in parser

azz:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no
| New feature?      | yes
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

Adding the parser for the nullish coalescing operator, `a ?? b`, currently at stage 1.

In the AST it isn an operator under `BinaryExpression`. Should it move to `LogicalExpression`?

Picked the plugin name: `"nullishCoalescingOperator"`. It's a bit of a mouthful, and may well change.

Ref: https://github.com/babel/proposals/issues/14 
Proposal: https://github.com/gisenberg/proposal-null-coalescing

Proposal will likely be renamed from `nullary` to `nullish`, see https://github.com/gisenberg/proposal-null-coalescing/issues/3.

I feel like I haven't added enough tests, but I'm not too sure of the edge cases I should test, let me know if you know of any!

PR #762, opened 2017-10-15
Add more tests for the nullish coalescing operator

azz:
| Q                 | A
| ----------------- | ---
| Bug fix?          | no
| Breaking change?  | no (breaks beta 28)
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | https://github.com/babel/babylon/pull/761#issuecomment-336668141
| License           | MIT

Added some more tests for associativity, multiline parsing, and precedence.

Due to this operator being similar to `||`, and it being in the `LogicalORExpression` non-terminal in the grammar, should it move from `BinaryExpression` to the `LogicalExpression` AST node?

Update:

`??` has been moved to `LogicalExpression`


hzoo:
Oh yeah that may be better actually

azz:
Has been changed from `Binary`- to `LogicalExpression` in this PR.

jridgewell:
So:

```js
a || b ?? c === (a || b) ?? c
a ?? b || c === (a ?? b) || c
```

But:
```js
a && b ?? c === (a && b) ?? c
a ?? b && c === a ?? (b && c)
```


Do we have any data for why this was chosen?

littledan:
@jridgewell Is that what this code does? In the spec I wrote for `??`, it parses *exactly* like `||`, and in particular `&&` groups more tightly.

jridgewell:
> Is that what this code does? 

Yes.

>  In the spec I wrote for ??, it parses exactly like ||, and in particular && groups more tightly.

That appears to be what's happening? I was just curious why it's not tighter than `&&` or looser than `||`.

littledan:
@jridgewell I thought it would make sense to be exactly the same because people are likely to be upgrading their `||` to `??`, and why change anything? If I were to change it, it would be to be looser than `||`. Seems like this deserves a bug to discuss further.

azz:
My intuition is it should be completely hot-swappable with `||`. Great for codemods/fixers.

littledan:
Opened https://github.com/tc39-transfer/proposal-nullish-coalescing/issues/15 to discuss this question; for now, the spec is hot-swappable like that.

Issue #756, opened 2017-10-11
Unable to parse a class get method with a decorator


mbehzad:
The error is thrown when the jsdoc lib is trying to parse a class with get method which has a decorator.
### Input Code

```js
  @override
  static get COMPONENT_NAME() {
    return "Overlay";
  }
```

### Babylon/Babel Configuration
used by the jsdoc lib

```js
{
    allowImportExportEverywhere: true,
    allowReturnOutsideFunction: true,
    allowSuperOutsideMethod: true,
    plugins: [
        'asyncGenerators',
        'bigInt',
        'classPrivateProperties',
        'classProperties',
        'decorators2',
        'doExpressions',
        'dynamicImport',
        'estree',
        'exportExtensions',
        'functionBind',
        'functionSent',
        'importMeta',
        'jsx',
        'numericSeparator',
        'objectRestSpread',
        'optionalCatchBinding',
        'optionalChaining'
    ],
    ranges: true,
    sourceType: env.conf.sourceType
}
```

### Expected Behavior
Should (?) parse a get method with decorator similar to a normal instance method.

### Current Behavior
Parser throws `ERROR: Unable to parse <file-path>: Stage 2 decorators may only be used with a class or a class method` error.

### Your Environment

| software         | version
| ---------------- | -------
| Babylon          |  7.0.0-beta.17
| node             | v7.8.0
| npm              | 3.10.8
| Operating System | windows 10 (64bit)

hzoo:
Hey @mbehzad! We really appreciate you taking the time to report an issue. The collaborators
on this project attempt to help as many people as possible, but we're a limited number of volunteers,
so it's possible this won't be addressed swiftly.

If you need any help, or just have general Babel or JavaScript questions, we have a vibrant Slack
community that typically always has someone willing to help. You can sign-up [here](https://slack.babeljs.io/)
for an invite.

PR #766, opened 2017-10-19
white lists getter and setter acceseors for having stage-2 decorators (#756)

mbehzad:
- adds `get` and `set` kind in addition to `method` to the list of allowed class members for having a decorator,
- adds tests for this two cases (decorator + set and decorator + get)

| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes/no
| Tests added/pass? | yes
| Fixed tickets     | #756
| License           | MIT

For stage 2 decorators, the `parseClassBody` checks if the decorator is attached to a class method. aka its `kind` is `method`. And throws an error if not.
commit af08e92 adds `set` and `get` to the class members which are allowed to have a decorator. (extends the check to `kind === set || get` in addition to `method`

### DIFF
--- a/src/parser/statement.js
+++ b/src/parser/statement.js
@@ -893,7 +893,7 @@ export default class StatementParser extends ExpressionParser {
 
       if (
         this.hasPlugin("decorators2") &&
-        member.kind != "method" &&
+        ["method", "get", "set"].indexOf(member.kind) === -1 &&
         member.decorators &&
         member.decorators.length > 0
       ) {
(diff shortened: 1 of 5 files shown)


mbehzad:
I don't know why the travis build failed. the tests work locally fine. Except the linting which throws error in types.js which was last changed with the usage of prettier in commit 5180ecdc (#600) couple of months ago. 🤔

danez:
Thank you very much for your first pull request.

PR #767, opened 2017-10-22
Fix decorators2 to support `export @decorator class A {}`

danez:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| Breaking change?  | no
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes
| Fixed tickets     | 
| License           | MIT

```js
export @decorator class A {}
```

I hope this is correct and should be parsable, but I guess it should.

Also added tests for the old decorators plugin, just to ensure it works there (works with different syntax though)


src/parser/statement.js
```diff
@@ -1418,7 +1418,8 @@ export default class StatementParser extends ExpressionParser {
       this.state.type.keyword === "let" ||
       this.state.type.keyword === "function" ||
       this.state.type.keyword === "class" ||
-      this.isContextual("async")
+      this.isContextual("async") ||
+      (this.hasPlugin("decorators2") && this.match(tt.at))
```
nicolo-ribaudo: Can we use `this.match(tt.at) && this.expectPlugin("decorators2")`?

littledan:
I agree that this should be supported. The [current spec text](https://tc39.github.io/proposal-unified-class-features/) supports it as well. Make sure to test decorated anonymous default class exports, which are specifically also supported. cc @wycats @bterlson

src/parser/statement.js
danez: okay changed

src/options.js
```diff
@@ -54,7 +54,7 @@ export const defaultOptions: Options = {
 export function getOptions(opts: ?Options): Options {
   const options: any = {};
   for (const key in defaultOptions) {
-    options[key] = opts && key in opts ? opts[key] : defaultOptions[key];
+    options[key] = opts && opts[key] != null ? opts[key] : defaultOptions[key];
```
danez: We have been accepting options even though they were `null` or `undefined`, and setting `plugins` to `null` for example did break the parser.

PR #773, opened 2017-10-27
Update parsing flow type/typeof imports

existentialism:
| Q                 | A
| ----------------- | ---
| Bug fix?          | y
| Breaking change?  | sorta?
| New feature?      | no
| Deprecations?     | no
| Spec compliancy?  | yes
| Tests added/pass? | yes/yes
| Fixed tickets     | 
| License           | MIT

Follow up to https://github.com/babel/babylon/pull/770, import type/typeof now matches the behavior in https://github.com/facebook/flow/commit/11eb1f997c1163146bbe8d127a43b2d071566186.

One test remains on the whitelist since [its comment](https://github.com/facebook/flow/blob/master/src/parser/test/flow/types/import_types/typeof_named_reserved_type.js) seems to conflict with its fixture.

/cc: @babel/flow


test/fixtures/flow/type-imports/invalid-import-type-3/actual.js
```diff
@@ -0,0 +1 @@
+import type { string } from "foo";
```
danez: duplicate test

test/fixtures/flow/type-imports/invalid-import-type-shorthand-4/actual.js
```diff
@@ -0,0 +1 @@
+import { type string } from "foo";
```
danez: Duplicate test

mroch:
@existentialism thanks for the cc. you're right about the busted test, it was a parser bug. fixed in https://github.com/facebook/flow/commit/a3e936a50db40dce3d44fc164369a1cc61a2e368

existentialism:
@mroch sorry I didn't report it sooner, was on my TODO list 😬

